PR #20, opened 2019-03-12
Update action resolution when can't find command line tools

MattKiazyk:
I had trouble with the proper syntax of setting Xcode-select via the command line - so included instructions on how to do it old-school with the mouse! 

My grade 7 Word Processing teacher would be ashamed of me!


interstateone:
Thanks Matt, I didn't know that you could do it in the preferences too. TIL!

PR #25, opened 2019-03-31
Saves the username and password of the iTunes account into the keychain

### LINKED ISSUE #11: Store password in keychain after it's entered

MattKiazyk:
Fixes #11 

Using KeychainAccess as a dependency. This automatically saves the username and password into the keychain for use at a later time.

Currently checks for Environment variables first, then keychain, then prompts the user. There is an awkward prompt when saving, with the keychain name that was set. Perhaps this should be set to more descriptive, and less syntactically correct.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -22,7 +22,7 @@ let package = Package(
         .target(
             name: "xcodes",
             dependencies: [
-                "Guaka", "XcodesKit"
+                "Guaka", "XcodesKit", "KeychainAccess"
             ]),
         .testTarget(
             name: "xcodesTests",
@@ -30,7 +30,7 @@ let package = Package(
         .target(
             name: "XcodesKit",
             dependencies: [
-                "AppleAPI", "Path", "Version", "PromiseKit", "PMKFoundation", "SwiftSoup", "LegibleError", "KeychainAccess"
+                "AppleAPI", "Path", "Version", "PromiseKit", "PMKFoundation", "SwiftSoup", "LegibleError"
             ]),
         .testTarget(
             name: "XcodesKitTests",
--- a/Package.swift
+++ b/Package.swift
@@ -15,7 +15,8 @@ let package = Package(
         .package(url: "https://github.com/mxcl/PromiseKit.git", .upToNextMinor(from: "6.8.3")),
         .package(url: "https://github.com/PromiseKit/Foundation.git", .upToNextMinor(from: "3.3.1")),
         .package(url: "https://github.com/scinfu/SwiftSoup.git", .upToNextMinor(from: "1.7.5")),
-        .package(url: "https://github.com/mxcl/LegibleError.git", .upToNextMinor(from: "1.0.1"))
+        .package(url: "https://github.com/mxcl/LegibleError.git", .upToNextMinor(from: "1.0.1")),
+        .package(url: "https://github.com/kishikawakatsumi/KeychainAccess.git", .upToNextMinor(from: "3.1.2"))
     ],
     targets: [
         .target(
@@ -29,7 +30,7 @@ let package = Package(
         .target(
             name: "XcodesKit",
             dependencies: [
-                "AppleAPI", "Path", "Version", "PromiseKit", "PMKFoundation", "SwiftSoup", "LegibleError"
+                "AppleAPI", "Path", "Version", "PromiseKit", "PMKFoundation", "SwiftSoup", "LegibleError", "KeychainAccess"
             ]),
         .testTarget(
             name: "XcodesKitTests",
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ error: terminated(72): xcrun --sdk macosx --find xctest output:
 
 ```
 
-If that occurs, it means you need to select a version of Xcode. You can do this with `xcode-select` or by choosing a Command Line Tools option in Xcode's preferences Locations tab. 
+If that occurs, it means you need to select a version of Xcode. You can do this with `xcode-select` or by choosing a Command Line Tools option in Xcode's preferences Locations tab.
 </details>
 
 ## Usage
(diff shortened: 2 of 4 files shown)


Sources/xcodes/main.swift
```diff
@@ -21,11 +26,41 @@ func loginIfNeeded() -> Promise<Void> {
...
+        return username
+    }
+    return nil
+}
+
+func findPassword() -> String? {
+    if let password = env(xcodesPassword) {
+        return password
+    }
+    else if let password = try? keychain.getString(xcodesPassword){
```
interstateone: This (and the equivalent code for the username) should use the same key as is used to set the username and password in the Keychain after a successful login, right?

Sources/xcodes/main.swift
```diff
@@ -21,11 +26,41 @@ func loginIfNeeded() -> Promise<Void> {
...
+    return nil
+}
+
+func login(_ username: String, password: String) -> Promise<Void> {
+    return firstly { () -> Promise<Void> in
+        manager.client.login(accountName: username, password: password)
+    }
+    .get { _ in
+        keychain["xcodesUsername"] = username
+        keychain["xcodesPassword"] = password
```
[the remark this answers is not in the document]
MattKiazyk: oops `keychain["xcodesUsername"] ` shouldn't have quotes - so both the save and load with the same property name
```
let xcodesUsername = "XCODES_USERNAME"
let xcodesPassword = "XCODES_PASSWORD"
```

I'm not sure what you're meaning by 
>  think it might be more appropriate to create a single application password item with the account set to the provided Apple ID and its password set to the provided password

I was setting 2 different keychain items (username/password).  The app will first check for any environment variables (for username/password) passed in, then go to the keychain, otherwise prompt. Are you thinking some other way?

Sources/xcodes/main.swift
MattKiazyk: As per my above comment `keychain["xcodesUsername"]` was wrong in setting it, so yes they should be using the same key

MattKiazyk:
> Can you elaborate which prompt this is? If I lock my login Keychain and try to run xcodes I'll see this, but I don't think we have much control over its copy:

Yeah it was that prompt - doesn't really give any context on what it's doing, just trying to access the keychain, but if there's nothing we can do there then ¯\_(ツ)_/¯

Sources/xcodes/main.swift
```diff
@@ -5,8 +5,13 @@ import PromiseKit
 import XcodesKit
 import LegibleError
 import Path
+import KeychainAccess
 
 let manager = XcodeManager()
+let keychain = Keychain(service: "com.interstateone.xcodes")
```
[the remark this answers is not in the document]
MattKiazyk: Fixed

Package.swift
```diff
@@ -29,7 +30,7 @@ let package = Package(
         .target(
             name: "XcodesKit",
             dependencies: [
-                "AppleAPI", "Path", "Version", "PromiseKit", "PMKFoundation", "SwiftSoup", "LegibleError"
+                "AppleAPI", "Path", "Version", "PromiseKit", "PMKFoundation", "SwiftSoup", "LegibleError", "KeychainAccess"
```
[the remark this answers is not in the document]
MattKiazyk: Moved

Sources/xcodes/main.swift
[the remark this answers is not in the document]
interstateone: Yes, instead of creating two keychain items and storing the username and password separately, create one and instead of setting the username to `XCODES_USERNAME`, set it it to the user's Apple ID and set the password to the password they entered.

README.md
```diff
@@ -40,6 +40,8 @@ E.g. `xcodes install 10.1`
 
 You'll then be prompted to enter your Apple ID username and password. You can also provide these with the `XCODES_USERNAME` and `XCODES_PASSWORD` environment variables.
 
+After successful log in, xcodes will save your Apple ID password into the keychain for quick retrieval subsequent tries.
```
interstateone:
```suggestion
After successful login, xcodes will save your Apple ID password into the keychain for quick retrieval on subsequent tries.
```

Sources/xcodes/main.swift
```diff
@@ -21,11 +26,37 @@ func loginIfNeeded() -> Promise<Void> {
     }
     .recover { error -> Promise<Void> in
         guard
-            let username = env("XCODES_USERNAME") ?? readLine(prompt: "Apple ID: "),
-            let password = env("XCODES_PASSWORD") ?? readSecureLine(prompt: "Apple ID Password: ")
+            let username = findUsername() ?? readLine(prompt: "Apple ID: "),
+            let password = findPassword(withUserName: username) ?? readSecureLine(prompt: "Apple ID Password: ")
```
interstateone:
```suggestion
            let password = findPassword(withUsername: username) ?? readSecureLine(prompt: "Apple ID Password: ")
```

Sources/xcodes/main.swift
```diff
@@ -21,11 +26,37 @@ func loginIfNeeded() -> Promise<Void> {
...
+}
+
+func findUsername() -> String? {
+    if let username = env(xcodesUsername) {
+        return username
+    }
+    return nil
+}
+
+func findPassword(withUserName username: String) -> String? {
```
interstateone:
```suggestion
func findPassword(withUsername username: String) -> String? {
```

Sources/xcodes/main.swift
```diff
@@ -21,11 +26,37 @@ func loginIfNeeded() -> Promise<Void> {
...
+        return password
+    }
+    return nil
+}
+
+func login(_ username: String, password: String) -> Promise<Void> {
+    return firstly { () -> Promise<Void> in
+        manager.client.login(accountName: username, password: password)
+    }
+    .get { _ in
```
interstateone:
```suggestion
    .done { _ in
```

interstateone:
Merged with https://github.com/interstateone/xcodes/commit/a7a073855ab427c2ab9146074562427396d3fdb7, thanks again Matt!

PR #32, opened 2019-04-26
Remove keychain if failed to log in #28

### LINKED ISSUE #28: Remove keychain item and prompt for password when login fails with a password from the keychain

MattKiazyk:
Resolves #28 

This catches any log in error with apple and removes the keychain password so that we don't try to use the keychain next time around.

### DIFF
--- a/Sources/xcodes/main.swift
+++ b/Sources/xcodes/main.swift
@@ -6,6 +6,7 @@ import XcodesKit
 import LegibleError
 import Path
 import KeychainAccess
+import AppleAPI
 
 let manager = XcodeManager()
 let keychain = Keychain(service: "ca.brandonevans.xcodes")
@@ -35,8 +36,7 @@ func loginIfNeeded(withUsername existingUsername: String? = nil) -> Promise<Void
         }
         .recover { error -> Promise<Void> in
             print(error.legibleLocalizedDescription)
-            // try again
-            print("Retry your password again")
+            print("Try entering your password again")
             return loginIfNeeded(withUsername: username)
         }
     }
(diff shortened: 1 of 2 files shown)


interstateone:
This looks like it'll do part of #28, but the other part I was hoping would happen was that after the incorrect password was removed from the keychain the user would then be prompted for their password as if they had originally logged in without a password in the keychain. Would you be able to add that as well?

MattKiazyk:
@interstateone not sure I fully undestand what you are looking for. Are you looking to basically call `readSecureLine` again to get a new password?

interstateone:
Yep, that's right. It would also be nice to have both "login without a saved password successfully" and "login with a saved password, fail, prompt for password and login successfully" reuse the same code where appropriate. I haven't put any thought into the details of that though.

MattKiazyk:
@interstateone i added in a retry for password. Currently there is not limit and it will continue to prompt for password. User can obviously cmd+c to quit to stop the loop. I thought that would be better then prompting to retry?

Sources/xcodes/main.swift
```diff
@@ -20,17 +21,24 @@ enum Error: Swift.Error {
...
         else { throw Error.missingUsernameOrPassword }
 
-        return login(username, password: password)
+        return firstly { () -> Promise<Void> in
+            login(username, password: password)
+        }
+        .recover { error -> Promise<Void> in
+            print(error.legibleLocalizedDescription)
+            print("Try entering your password again")
+            return loginIfNeeded(withUsername: username)
```
interstateone: When I was testing to make sure that the keychain item was only removed when there was a password issue, I found that if I turned off my internet connection it would get stuck in a retry loop that never prompted for a password. I think it didn't prompt because the password was still in the keychain (and returned in `findPassword` above) but it'll never succeed because the internet isn't available. Seems like the same check for a password failure should happen here too.

I'm sorry that I didn't find this sooner during review. 😞 I think this'll be the last thing before merging.

```suggestion
            if case Client.Error.invalidUsernameOrPassword = error {
                print("Try entering your password again")
                return loginIfNeeded(withUsername: username)
            }
            else {
                return Promise(error: error)
            }
```

interstateone:
Merged with [suggested changes](https://github.com/RobotsAndPencils/xcodes/pull/32#discussion_r282721488) in c12b7c4939202fe8b5e1fa38ee578ed7beeb1d65

PR #44, opened 2019-05-16
Update SwiftSoup to 2.0

MattKiazyk:
Updates SwiftSoup dependency to v 2.0

This updates warnings that were visible through xcode. As well as supports Swift 5 - which should help in #31


MattKiazyk:
Closing as I see this is updated in #31 already!

PR #46, opened 2019-05-16
Add a new configuration file and save username into it after a successful login

### LINKED ISSUE #40: [Feature request] Store username in config/cache after every successful login
Currently, users have to enter their Apple ID username or specify `XCODES_USERNAME` environment variable every time they run the `xcodes update` or `xcodes install` commands (and also `xcodes list` for the first time). This feature will eliminate the need of asking users to type in their username every time.

## Implementation details
- If a username isn't specified with the `XCODES_USERNAME` environment variable, and there is no
username saved, prompt the user for a username.
  - If the login succeeds, then save the username.
  - If the login fails, don't save the username.
- If a username isn't specified with the `XCODES_USERNAME` environment variable, and there is a username saved, use that username for login and don't prompt the user for a username.
- If a username is specified with the `XCODES_USERNAME` environment variable, use that username for login (don't check for a saved username, don't prompt the user for a username).
  - If the login succeeds, then save the username.
  - If the login fails, don't save the username.

thii:
Resolves #40

### DIFF
--- a/Sources/XcodesKit/Models.swift
+++ b/Sources/XcodesKit/Models.swift
@@ -51,3 +51,7 @@ public struct InfoPlist: Decodable {
         case bundleShortVersion = "CFBundleShortVersionString"
     }
 }
+
+public struct Configuration: Codable {
+    public let defaultUsername: String?
+}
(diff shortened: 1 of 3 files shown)


interstateone:
Merged in ddd48f2371dd69623f0542c09ffe1b5d755a1262 with small name change for consistency. Thanks @thii!

PR #53, opened 2019-06-18
Refactor to better define boundaries

interstateone:
XcodeManager had a lot of responsibilities and a poor name, so this PR intends to fix that. XcodeManager is now XcodeList, and provides arrays of available and installed Xcodes. This better differentiates it from XcodeInstaller, which focuses on the download and install process. The Configuration type and some extensions have been moved into more appropriate places too.

**How to test:**

It should be sufficient to make sure that the `xcodes list`, `xcodes installed`, and `xcodes update` commands continue to work as expected. You can run these with `swift run COMMAND`.

### DIFF
--- a/Sources/XcodesKit/Path+.swift
+++ b/Sources/XcodesKit/Path+.swift
@@ -1,0 +1,8 @@
+import Path
+
+extension Path {
+    static let oldXcodesApplicationSupport = Path.applicationSupport/"ca.brandonevans.xcodes"
+    static let xcodesApplicationSupport = Path.applicationSupport/"com.robotsandpencils.xcodes"
+    static let cacheFile = xcodesApplicationSupport/"available-xcodes.json"
+    static let configurationFile = xcodesApplicationSupport/"configuration.json"
+}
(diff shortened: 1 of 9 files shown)


provMattKiazyk:
Test

Sources/XcodesKit/Path+.swift
```diff
@@ -0,0 +1,8 @@
+import Path
+
+extension Path {
+    static let oldXcodesApplicationSupport = Path.applicationSupport/"ca.brandonevans.xcodes"
```
MattKiazyk: This is probably new to me but what is with the quotes?

MattKiazyk:
Another thing - is with the refactoring, it no longer opens in xcode 11 properly

Sources/XcodesKit/Path+.swift
interstateone: The `/` is an infix operator that does the same thing as `Path.join`, so the quotes here are just because `ca.brandonevans.xcodes` is a String literal. It would be equivalent to `Path.applicationSupport.join("ca.brandonevans.xcodes")`.

PR #55, opened 2019-06-21
Improve prerelease version comparison

### LINKED ISSUE #22: Improve pre-release version comparison
See release [0.3.1](https://github.com/interstateone/xcodes/releases/tag/0.3.1) and acd05f937fe9f79afa71c936643396b63f9ded81. Ideally the output of `xcodes installed` and `xcodes list` would be more precise to reflect the installed and available pre-release versions. Need to check if there's enough info available before downloading a pre-release version and in an installed Info.plist to do this.

Pre-release builds publish a product version on developer.apple.com/downloads, e.g. Xcode 11 Beta 1 is `11M336W`. The app bundle contains a file at Contents/version.plist and the value for ProductBuildVersion matches this, although it should be compared case-insensitively because it is `11M336w`.

### LINKED ISSUE #54: Different beta versions installed to the same location
When I installed Xcode 11 beta 2 it replaced the previously installed Xcode 11 beta 1 at `/Applications/Xcode-11.0.0-beta`. (`xcode-install` does keep each beta version separately)

It would be nice to keep the previously installed beta versions as is. The old betas are removed from the Apple website when the new beta is available, and sometimes there are new issues with the latest beta so being able to switch back would be nice (happened to us last year with Xcode 10 beta 5).

interstateone:
In order to improve the list of all available Xcodes and which ones are installed, we need to improve how Versions are compared.

We can only know the build number of released Xcodes once they're installed, because this isn't provided by the /downloads/more data. Beta releases only have prerelease identifiers on the /downloads page, but we can get build numbers on that same page and once downloaded. So for released Xcodes, we'll only compare the major, minor and patch values. For beta releases, we'll compare the major, minor, patch and build number, ignoring the prerelease identifier (e.g. Beta 2).

The list should include the prerelease identifiers for beta releases if we can find it, and so we'll try to parse that from the filename (in case it was installed with xcodes) or default to simply "beta".

Resolves #22 
Resolves #54

Example output using xcodes built from this branch:

```
$ xcodes list
10.0
10.1 (Installed)
10.2
10.2.1 (Installed)
11.0 Beta (11M336w) (Installed)
11.0 Beta 2 (11M337n)

$ xcodes install 11.0 beta 2
Downloading Xcode 11.0.0-beta.2+11M337n: 100%
Xcode 11.0.0-beta.2+11M337n failed its security assessment with the following output:

/Applications/Xcode-11.0.0-beta.2.app: rejected (invalid destination for symbolic link in bundle)

It remains installed at /Applications/Xcode-11.0.0-beta.2.app if you wish to use it anyways.

$ xcodes list
...
10.0
10.1 (Installed)
10.2
10.2.1 (Installed)
11.0 Beta (11M336w) (Installed)
11.0 Beta 2 (11M337n) (Installed)
```

Tested again today:
```
$ .build/release/xcodes list
...
10.2.1 (Installed)
11.0 Beta (11M336w) (Installed)
11.0 Beta 2 (11M337n) (Installed)
11.0 Beta 3 (11M362V)

$ .build/release/xcodes install 11 beta 3
...
Downloading Xcode 11.0.0-beta.3+11M362V: 8%
```

Things to test:

- Does it correctly list the installed state with `xcodes list` when you only have beta 1 installed?
- Does it correctly list the installed state with `xcodes list` when you have beta 1 and 2 installed?
- Can you install beta 2 without overwriting beta 1?


interstateone:
Made one additional change in order to perform a case-insensitive comparison of build metadata identifiers because beta 3 was 11M362V online but 11M362v on disk.

PR #68, opened 2019-08-30
Replace CircleCI with GitHub Actions

interstateone:
This removes the CircleCI config and badge and adds a workflow to run tests on GitHub's macOS runners. Pretty straightforward conversion as far as config goes, and tests are passing: https://github.com/RobotsAndPencils/xcodes/runs/207600895. [From what I've seen](https://github.com/badges/shields/issues/2574#issuecomment-524353429) the badge will only show once this config is merged into master.


daveanderson:
👏

PR #77, opened 2019-10-04
Use Swift 5.1 and Xcode 11

### LINKED ISSUE #75: Swift 5.1 and Xcode 11
Resolving this should:

- [x] perform only enough code changes to make things work, silence warnings, etc.
- [x] Update .gitignore so we can commit the .swiftpm directory without shared user data
- [x] Remove .vscode directory
- [x] Update .swift-version file
- [x] Update Package.swift
- [x] CI workflow should be run using Xcode 11

This scope doesn't include a rewrite to use Combine or anything like that.

interstateone:
See individual commits, resolves #75.

Notably, building with optimizations (so, release builds) gets stuck building Guaka. This appears to be related to [this bug](https://bugs.swift.org/browse/SR-11010), and a fix has landed [here](https://github.com/apple/swift/pull/26660) which as of now is in the Swift 5.1 development snapshot toolchains. So two workarounds are to use a prerelease toolchain or build with `-Onone`. 🤞 the fix ships in an Xcode soon (it's not in 11.2 beta).

I opened FB7347879 with this project attached to track and help solve the problem.


daveanderson:
👀

daveanderson:
Xcode 11.2 doesn't or does not include the fix?

interstateone:
> Xcode 11.2 doesn't or does not include the fix?

The fix isn't included in the Swift toolchain shipped with 11.2 beta 1.

> I don't see the optimization setting set

Thanks for reminding me, I wanted to file a problem report before adding that. Done now!

PR #84, opened 2019-11-13
Update Apple auth URLs

### LINKED ISSUE #83: Apple Developer authentication has changed
At some point today (2019/11/12) Apple changed how their auth works. This involves not just a small change like the URL `https://olympus.itunes.apple.com/v1/session` to `https://appstoreconnect.apple.com/olympus/v1/session` but a larger one like where the `X-Apple-Widget-Key` header value comes from. Will need to do some more digging to find out how we can replicate it.

interstateone:
The URLs we were using were originally from the xcode-install project and had been working up until today (2019/11/12). These URLs [were updated](https://github.com/fastlane/fastlane/pull/14509) in that project (actually in fastlane, which it uses for auth) about 7 months ago but I didn't notice this. I started to look in a web inspector while signing in to developer.apple.com and saw that the Olympus session URL changed. [Dave DeLong noted](https://github.com/RobotsAndPencils/xcodes/issues/83#issuecomment-553207943) that a slightly different URL worked for the iTC service key, which is also what [xcinfo uses](https://github.com/xcodereleases/xcinfo/blob/e51f022db3584be884b5d10d8b161ac26dbbee0b/Sources/OlympUs/OlympUs.swift#L213).

**Testing**

1. Delete the `com.robotsandpencils.xcodes` keychain item
1. `swift run xcodes update` should complete by printing out all the available Xcode versions which as of today should end with `11.2.1`

Resolves #83

### DIFF
--- a/Sources/AppleAPI/URLRequest+Apple.swift
+++ b/Sources/AppleAPI/URLRequest+Apple.swift
@@ -1,12 +1,12 @@
 import Foundation
 
 extension URL {
-    static let itcServiceKey = URL(string: "https://olympus.itunes.apple.com/v1/app/config?hostname=itunesconnect.apple.com")!
+    static let itcServiceKey = URL(string: "https://appstoreconnect.apple.com/olympus/v1/app/config?hostname=itunesconnect.apple.com")!
     static let signIn = URL(string: "https://idmsa.apple.com/appleauth/auth/signin")!
     static let authOptions = URL(string: "https://idmsa.apple.com/appleauth/auth")!
     static let submitSecurityCode = URL(string: "https://idmsa.apple.com/appleauth/auth/verify/trusteddevice/securitycode")!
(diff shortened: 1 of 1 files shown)


MattKiazyk:
Note: I didn't delete the keychain item before - but it did prompt me for keychain access again. After typing in my password nothing happened on the console. `xcodes update` then connected  properly and printed out the correct versions. 

```
9.4
9.4.1
10.0
10.1
10.2 (Installed)
10.2.1
10.3
11.0
11.1 GM Seed (11A1027) (Installed)
11.1
11.2
11.2.1
```

interstateone:
Adding to the record from a side conversation: prompting for keychain access is expected sometimes, but not printing _anything_ (error or list) is not expected. I can't reproduce locally right now but will note for a future fix.

PR #94, opened 2020-06-24
Replace Guaka with swift-argument-parser

interstateone:
I have a slight preference for using the first-party library for this because it seems to be more actively developed.

This PR was a draft for a long time because I was waiting to see if SAP would implement fuzzy command suggestions like Guaka has. (SAP does have this for options.) But I'd rather get this merged instead of constantly rebasing it on other changes, and yank the bandaid off. I don't think the lack of this feature is a dealbreaker for adoption, and it might even motivate me to (try to) implement it. 🙃 

What does this change?

- shell completions are supported
- help text has different formatting than what Guaka printed
- fuzzy command suggestions aren't implemented in SAP yet, so errors when you mistype a command are unhelpful

This doesn't change any of the commands, options, flags, or other behaviour.

## Testing

Exploratory testing is probably best here, try all the commands and make sure they still work the same way.

## Help output

To get a high level diff of how this change appears to users, I made a table with the help output before and after. This didn't really work well as a markdown table so I've attached a PDF of it: [help output.pdf](https://github.com/RobotsAndPencils/xcodes/files/5768080/help.output.pdf)

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,17 +6,8 @@
         "repositoryURL": "https://github.com/PromiseKit/Foundation.git",
         "state": {
           "branch": null,
-          "revision": "0a59b0d89f469979aecacdf9046df14f99d1cc6b",
-          "version": "3.3.2"
-        }
-      },
-      {
-        "package": "Guaka",
-        "repositoryURL": "https://github.com/nsomar/Guaka.git",
-        "state": {
-          "branch": null,
-          "revision": "6fb29b2378166a30d72120980e1c099c664598de",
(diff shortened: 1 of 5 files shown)


csykes (approved):
I'm a Fish shell guy so no issues when I ran it from there (but I couldn't get shell complete to work). The documentation conversion is more complete so that's a nice side effect.

interstateone:
Hmm, okay, that's good to know. I had a heck of a time getting it to work in zsh too, but I don't know if that had anything to do with my years of accreted config files.

PR #107, opened 2020-10-02
Add support for 2FA accounts using SMS

### LINKED ISSUE #96: Login with SMS 2FA code fails with "The data couldn’t be read because it is missing"
Can't get the list of Xcodes from Apple when login with 2FA account.

```
$ xcodes list
Apple ID: foo@bar.com
Apple ID Password:
Enter the code:
000000
The data couldn’t be read because it is missing.
```

Second try is the same:
```
$ xcodes list
Enter the code:
000001
The data couldn’t be read because it is missing.
```

interstateone:
Adds support for 2FA accounts that use SMS instead of trusted devices for delivery. This should close #96, but I'll need to get confirmation with the folks that reported that issue.

There are a couple 2FA flows that are now supported:

1. 2FA SMS automatically sent because user has a single phone number
1. 2FA that requires selection from multiple phone numbers before sending an SMS
1. 2FA automatically displayed on the screen of trusted devices

I'm deliberately not supporting the earlier two-step authentication because upgrading the account to 2FA seems like a reasonable workaround.

Some tests have been added that verify the flow handles different situations correctly, and prints the correct copy to the terminal.

## Testing

I've tested this myself with:

- an account that has no trusted devices added and a single phone number
- my existing account that has multiple trusted devices added and a single phone number

The latter case was able to validate the selection flow in case 2 above by opting out of the trusted device code by entering `sms`, although there was only one phone number to choose from.

```
cd xcodes
git checkout -b interstateone-sms-2fa-support master
git pull https://github.com/interstateone/xcodes.git sms-2fa-support
```

You can test this new behaviour with your own account using a command like:

```
swift run xcodes update
```

If you need to change to use a different account than one that was already signed in, you can set the XCODES_USERNAME environment variable first:

```
XCODES_USERNAME=my_other_apple_id@example.com swift run xcodes update
```

If you already had a valid session in xcodes then you'll need to reset the cookies, cache, etc. in order to test this flow. You can add the following snippet in main.swift at line 15. (We may want to add a flag for this later, for edge cases or just debug/testing.)

```swift
let s = DispatchSemaphore(value: 0)
URLSession.shared.reset {
    print("URLSession has been reset")
    s.signal()
}
s.wait()
```

With all that said, what I'd expect to happen when you test this is for SMS 2FA to "just work" now. If it doesn't, or if you think some of the instructional copy that's shown during the flow could be improved, please let me know.

---

The Paw file has been updated to support testing the SMS flow.

In order to make it a bit easier (or at least familiar) to test the AppleAPI code I changed it to use the same Environment struct that XcodesKit does. I'd started out development using OHHTTPStubs but wanted a bit more, and this also allows testing other things like the log output more easily which is useful for validating longer interactions.

Updates to Swift 5.3 and Xcode 12, which allowed transitioning to use SPM resources for test fixtures.


Sources/AppleAPI/Client.swift
```diff
@@ -87,24 +90,209 @@ public class Client {
...
+    }
+    
+    func promptForSMSSecurityCode(length: Int, for trustedPhoneNumber: AuthOptionsResponse.TrustedPhoneNumber) -> SecurityCode {
+        let code = Current.shell.readLine("Enter the \(length) digit code sent to \(trustedPhoneNumber.numberWithDialCode): ") ?? ""
+        return .sms(code: code, phoneNumberId: trustedPhoneNumber.id)
+    }
+    
+    func handleWithPhoneNumberSelection(authOptions: AuthOptionsResponse, serviceKey: String, sessionID: String, scnt: String) -> Promise<Void> {
+        return firstly {
+            selectPhoneNumberInteractively(from: authOptions.trustedPhoneNumbers!)
```
aerickson14, line 194: is it guaranteed that there will always be trusted phone numbers?

interstateone: Good question, this'll now throw an appropriate error if trustedPhoneNumbers is nil or empty. I don't think this should be possible because [2FA requires at least one trusted phone number](https://support.apple.com/en-ca/HT204915), but can't hurt to check since this is all private-ish API.

aerickson14 (approved):
Tested this out with my account that uses SMS and it work well 👍

PR #110, opened 2020-10-15
Retry installation when it fails because of damaged XIP

### LINKED ISSUE #92: The archive is damaged and can’t be expanded.
Hi!
I seem to have run into a situation where xcodes chokes a bit.  I think what happened is that I closed my laptop while it was downloading and then the next time I opened it and ran this again I started getting this error from which it can't recover by itself.  I have ample free space so I think the archive was corrupt because it didn't finish.  In any case I worked around it by deleting the archive but it would be nice if in the case of corruption it would get it again.

Screenshot 2020-05-31 14 55 43

interstateone:
This addresses #92, where xcodes tries to use an existing XIP but then fails if it's somehow damaged. Now, it'll remove the damaged XIP and retry once. If the user provided the path to the XIP then it'll just exit with an error without removing the XIP, so the user can decide what to do with it. I don't think it makes sense to retry more than once, since I can't imagine e.g. downloading a damaged XIP twice and _then_ getting a good one.

A couple changes were made to support this:

- Break up the main `install` method to make it easier to retry
- Catch the Process.PMKError early and map it to a more specific error. This might be good to do for other shell errors, and would remove the need for handling Process.PMKError specially in the install command in main.swift

## Testing

A test was added to cover this case, and you can test it manually by creating a fake, empty XIP in the expected location.

```shell
# If you already have Xcode 12.0.0 installed, pick another available version that you don't
# Run these two commands, and you should see output similar to what's below
$ touch ~/Library/Application\ Support/com.robotsandpencils.xcodes/Xcode-12.0.0.xip
$ swift run xcodes install 12.0.0
(1/6) Found existing archive that will be used for installation at $HOME/Library/Application Support/com.robotsandpencils.xcodes/Xcode-12.0.0.xip.
The archive "Xcode-12.0.0.xip" is damaged and can't be expanded.
Removing damaged XIP and re-attempting installation.

(1/6) Downloading Xcode 12.0.0: 5%
# ...
```

Closes #92

### DIFF
--- a/Sources/XcodesKit/XcodeInstaller.swift
+++ b/Sources/XcodesKit/XcodeInstaller.swift
@@ -11,6 +11,7 @@ public final class XcodeInstaller {
     static let XcodeCertificateAuthority = ["Software Signing", "Apple Code Signing Certification Authority", "Apple Root CA"]
 
     public enum Error: LocalizedError, Equatable {
+        case damagedXIP(url: URL)
         case failedToMoveXcodeToApplications
         case failedSecurityAssessment(xcode: InstalledXcode, output: String)
         case codesignVerifyFailed(output: String)
@@ -27,6 +28,8 @@ public final class XcodeInstaller {
 
         public var errorDescription: String? {
             switch self {
+            case .damagedXIP(let url):
+                return "The archive \"\(url.lastPathComponent)\" is damaged and can't be expanded."
             case .failedToMoveXcodeToApplications:
                 return "Failed to move Xcode to the /Applications directory."
             case .failedSecurityAssessment(let xcode, let output):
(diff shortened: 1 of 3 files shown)


Sources/XcodesKit/XcodeInstaller.swift
```diff
@@ -132,6 +135,47 @@ public final class XcodeInstaller {
...
+        return firstly { () -> Promise<(Xcode, URL)> in
+            return self.getXcodeArchive(installationType)
+        }
+        .then { xcode, url -> Promise<InstalledXcode> in
+            return self.installArchivedXcode(xcode, at: url)
+        }
+        .recover { error -> Promise<InstalledXcode> in
+            switch error {
+            case XcodeInstaller.Error.damagedXIP(let damagedXIPURL):
+                guard attemptNumber <= 1 else { throw error }
```
civatrix, line 157: Won't this retry twice? When `attemptNumber` is 0 and 1?

civatrix (approved):
I think this will retry twice but otherwise looks fine

Sources/XcodesKit/XcodeInstaller.swift
interstateone: 😅 sure does, thank you :) changed to `< 1`

PR #111, opened 2020-10-16
Add support for downloading with aria2 if it's already installed

### LINKED ISSUE #93: Investigate performing downloads with aria2
- [x] Test out how much of an improvement it is on its own, if any
- [x] Prove out performing downloads with aria2. Can it work with no user-facing changes to xcodes? Assume that users are responsible for installing it separately (it's in homebrew) and xcodes will detect that and run it in a separate process.

Reference:
https://github.com/vineetchoudhary/Downloader-for-Apple-Developers
https://blog.kulman.sk/faster-way-to-download-and-install-xcode/

interstateone:
By default, xcodes will use a URLSession to download the specified version. If [aria2](https://aria2.github.io) (available in Homebrew with `brew install aria2`) is installed, either at /usr/local/bin/aria2c or at the path specified by the --aria2 flag, then it will be used instead. aria2 will use up to 16 connections to download Xcode 3-5x faster (based on a couple tests on my residential internet). If you have aria2 installed and would prefer to not use it, you can use the --no-aria2 flag.

aria2 downloads should retry up to 3 times if it fails in the middle, just like URLSession. Errors that cause aria2c to exit with a non-zero status code will show an appropriate error message.

In order to support both aria2 and URLSession, I had to change the download URLs. The previous URL prefix still works, but I was having issues with it when using aria2 for downloads. This new URL prefix matches whats shown in a browser at developer.apple.com/download/more, and works with both URLSession and aria2.

Closes #93 

## Testing

Download a bunch of xcodes while you're on this branch 😄 

```
cd xcodes
git checkout -b interstateone-aria2 master
git pull https://github.com/interstateone/xcodes.git aria2
```

You can test this new behaviour with your own account using a command like:

```
swift run xcodes install 12.1 gm seed # or some other version that you don't have already
```

You should try installing both with and without aria2, so:
- without aria2c
- with aria2c
- with aria2c and --no-aria2 flag to use URLSession

You can see if aria2c is being used during install by opening Activity Monitor and filtering for `aria2c` in the top right.

### DIFF
--- a/Sources/XcodesKit/Environment.swift
+++ b/Sources/XcodesKit/Environment.swift
@@ -53,6 +53,81 @@ public struct Shell {
     public func xcodeSelectSwitch(password: String?, path: String) -> Promise<ProcessOutput> {
         xcodeSelectSwitch(password, path)
     }
+    
+    public var downloadWithAria2: (Path, URL, Path, [HTTPCookie]) -> (Progress, Promise<Void>) = { aria2Path, url, destination, cookies in
+        let process = Process()
+        process.executableURL = aria2Path.url
+        process.arguments = [
+            "--header=Cookie: \(cookies.map { "\($0.name)=\($0.value)" }.joined(separator: "; "))",     
+            "--max-connection-per-server=16",
+            "--split=16",
+            "--summary-interval=1",
+            "--stop-with-process=\(ProcessInfo.processInfo.processIdentifier)",
+            "--dir=\(destination.parent.string)",
+            "--out=\(destination.basename())",
+            url.absoluteString,
+        ]
+        let stdOutPipe = Pipe()
+        process.standardOutput = stdOutPipe
+        let stdErrPipe = Pipe()
+        process.standardError = stdErrPipe
+        
+        var progress = Progress(totalUnitCount: 100)
+
+        let observer = NotificationCenter.default.addObserver(
+            forName: .NSFileHandleDataAvailable, 
+            object: nil, 
+            queue: OperationQueue.main
+        ) { note in
+            guard
+                // This should always be the case for Notification.Name.NSFileHandleDataAvailable
+                let handle = note.object as? FileHandle,
+                handle === stdOutPipe.fileHandleForReading || handle === stdErrPipe.fileHandleForReading
+            else { return }
+
+            defer { handle.waitForDataInBackgroundAndNotify() }
+
+            let string = String(decoding: handle.availableData, as: UTF8.self)
+            let regex = try! NSRegularExpression(pattern: #"((?<percent>\d+)%\))"#)
+            let range = NSRange(location: 0, length: string.utf16.count)
+
+            guard
+                let match = regex.firstMatch(in: string, options: [], range: range),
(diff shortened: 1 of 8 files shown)


thii:
Nice. It'd be nice to make `aria2` a dependency in the Homebrew formula so that it'd be using it by default.

interstateone:
@thii That’s a good idea, I hadn’t thought of that! I’m leaning towards making it optional in the formula, at least to begin with, so that this is an incremental change.

(Part of my hesitation is that installing aria2 took _hours_ despite providing a bottle, because it had to build a bunch of its dependencies. I think this might have been because of the versions of some formulas that I already had installed, and so not everyone will have this experience, but I’d like to see how it works for more people first.)

interstateone:
If we do want to make aria2 the default download option in the future I'd like to look into statically linking libaria2 instead of relying on the separate executable. I don't know if this would actually work yet, though 😄

thii:
I just noticed the installation slowness too. Let's keep it optional for now.

Sources/XcodesKit/Environment.swift
```diff
@@ -53,6 +53,75 @@ public struct Shell {
...
+        process.standardError = stdErrPipe
+        
+        var progress = Progress(totalUnitCount: 100)
+
+        let token = NotificationCenter.default.addObserver(
+            forName: .NSFileHandleDataAvailable, 
+            object: nil, 
+            queue: OperationQueue.main
+        ) { note in
+            let handle = note.object as! FileHandle
```
civatrix, line 82: I assume this `as!` is safe and doesn't need to be unwrapped?

Sources/XcodesKit/Environment.swift
```diff
@@ -53,6 +53,75 @@ public struct Shell {
...
+        
+        do {
+            try process.run()
+        } catch {
+            return (progress, Promise(error: error))
+        }
+
+        let promise = Promise<Void> { seal in
+            // Capture token 
+            _ = token
```
civatrix, line 107: I don't understand what this is doing.

civatrix (approved):
A few questions about the Aria2 code but otherwise looks good

Sources/XcodesKit/Environment.swift
interstateone: Yeah that's the case in the docs for `Notification.Name.NSFileHandleDataAvailable`, but I'll change it to return early if for some reason it isn't.

Sources/XcodesKit/Environment.swift
interstateone: Mmm, it's wrong anyways, thanks for making me check! 😄 The intent was to capture `token` to keep the observation alive while the promise was, but block-based observation still requires removing the observer (unlike selector-based observation.) I've changed this to remove the token after the promise is resolved.

civatrix (approved):

interstateone:
I tried adding `depends_on` to the formula, but it seems to only work when building from source, not when installing the bottle. I think it's okay if users have to `brew install robotsandpencils/made/xcodes aria2` instead though.

Issue #114, opened 2020-11-08
Allow selecting with short version


thii:
Consider:

```
$ xcodes installed
10.0
10.1
10.2
10.2.1
10.3
11.6
11.7 (Selected)
12.0 GM Seed (12A7209)
12.1 GM Seed (12A7403)
12.2 Release Candidate (12B5044c)

$ xcodes select 12.0
Available Xcode versions:
1) 10.0
2) 10.1
3) 10.2
4) 10.2.1
5) 10.3
6) 11.6
7) 11.7 (Selected)
8) 12.0 GM Seed (12A7209)
9) 12.1 GM Seed (12A7403)
10) 12.2 Release Candidate (12B5044c)
Enter the number of the Xcode to select:
```

Running `xcodes select 12.0` does not select `12.0 GM Seed (12A7209)` even though it was the only 12.0 installed.

Please allow using the short version with the `select` command when the provided version is not ambiguous.

PR #115, opened 2020-11-12
Allow selecting with short version

thii:
If a short version is provided, xcodes will select a matched one if
there is only a single match found in the machine. This allows, for
example, if user has Xcode 12.0 GM Seed installed and has no other 12.0
installed, `xcodes select 12.0` will select it.

Fixes https://github.com/RobotsAndPencils/xcodes/issues/114.

### DIFF
--- a/Sources/XcodesKit/Version+.swift
+++ b/Sources/XcodesKit/Version+.swift
@@ -8,6 +8,12 @@ public extension Version {
                prereleaseIdentifiers == other.prereleaseIdentifiers
     }
 
+    func isEqualWithoutAllIdentifiers(to other: Version) -> Bool {
+        return major == other.major &&
+               minor == other.minor &&
+               patch == other.patch
+    }
+
     /// If release versions, don't compare build metadata because that's not provided in the /downloads/more list
     /// if beta versions, compare build metadata because it's available in versions.plist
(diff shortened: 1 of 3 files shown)


interstateone (approved):
Thanks for taking the time to open the issue and submit this PR that implements it, @thii! This is definitely a nice improvement.

I think we could also implement this behaviour for the install ([here](https://github.com/RobotsAndPencils/xcodes/blob/master/Sources/XcodesKit/XcodeInstaller.swift#L262)) and uninstall ([here](https://github.com/RobotsAndPencils/xcodes/blob/master/Sources/XcodesKit/XcodeInstaller.swift#L482)) commands. I know in the past it was unclear to some users that you had to include prerelease identifiers in order to install betas (e.g. `12.3 beta` instead of simply `12.3`). Is this something you'd be interested in implementing?

*edit* I've opened #116 to track that.

thii:
I can take a look at it some time next week. In the meantime, would you mind cutting a release that includes this change? We're migrating from xcode-install and are depending on this behavior.

interstateone:
Okay. I'd prefer the next version to use this behaviour consistently across the commands, so I'll tackle #116 this weekend and make a new release soon.

PR #117, opened 2020-11-14
Allow somewhat fuzzy version matching for install and uninstall commands

### LINKED ISSUE #116: Allow installing and uninstalling prerelease versions without prerelease identifiers
Along the lines of #114, I think we can implement this for the install and uninstall commands as well. For example, when there is only a beta version of 12.3 available and it's unambiguous, this would allow `xcodes install 12.3` instead of requiring `xcodes install 12.3 beta`.

interstateone:
This PR extracts the behaviour added in #115, which allows users to omit prerelease identifiers (like "beta 2") in command arguments when it's unambiguous, and uses it in the install and uninstall commands. If the version is ambiguous then it should fail.

## Testing

```
cd xcodes
git checkout -b interstateone-116-fuzzy-install-uninstall master
git pull https://github.com/interstateone/xcodes.git 116-fuzzy-install-uninstall
```

Right now there's a 12.3 beta available, and with this change you should be able to install it without specifying the beta identifier:

```
swift run xcodes update
swift run xcodes install 12.3
```

The tests that were added provide examples of how this should work in other situations, and you can try out the install and uninstall commands to verify. If there's an exact match even if there are multiple possible matches (like `12.2`, when `12.2`, `12.2 beta 1`, `12.2 beta 2`, `12.2 beta 3`, and `12.2 release candidate` are available) then it'll match `12.2`. Once the next 12.3 beta is released then `12.3` is no longer unambiguous and would fail to match.

Closes #116.

### DIFF
--- a/Tests/XcodesKitTests/Models+FirstWithVersionTests.swift
+++ b/Tests/XcodesKitTests/Models+FirstWithVersionTests.swift
@@ -1,0 +1,104 @@
+import Path
+import XCTest
+import Version
+@testable import XcodesKit
+
+final class ModelsFirstWithVersionTests: XCTestCase {
+    let xcodes = [
+        Xcode(version: Version(xcodeVersion: "1.2.3")!,        url: URL(fileURLWithPath: "https://developer.apple.com/Xcode1.2.3.app"),      filename: "Xcode1.2.3.app",      releaseDate: nil),
+        Xcode(version: Version(xcodeVersion: "1.2.3 Beta 1")!, url: URL(fileURLWithPath: "https://developer.apple.com/Xcode1.2.3Beta1.app"), filename: "Xcode1.2.3Beta1.app", releaseDate: nil),
+        Xcode(version: Version(xcodeVersion: "1.2.3 Beta 2")!, url: URL(fileURLWithPath: "https://developer.apple.com/Xcode1.2.3Beta2.app"), filename: "Xcode1.2.3Beta2.app", releaseDate: nil),
+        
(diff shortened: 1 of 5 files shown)


thii:
Thanks for this.

Tests/XcodesKitTests/Models+FirstWithVersionTests.swift
```diff
@@ -0,0 +1,104 @@
...
+        Xcode(version: Version(xcodeVersion: "10.11.12 Release Candidate")!, url: URL(fileURLWithPath: "https://developer.apple.com/Xcode10.11.12ReleaseCandidate.app"), filename: "Xcode10.11.12ReleaseCandidate.app", releaseDate: nil),
+    ]
+    
+    let installedXcodes = [
+        InstalledXcode(path: Path("/Applications/Xcode-1.2.3.app")!,        version: Version(xcodeVersion: "1.2.3")!),
+        InstalledXcode(path: Path("/Applications/Xcode-1.2.3-beta.1.app")!, version: Version(xcodeVersion: "1.2.3 Beta 1")!),
+        InstalledXcode(path: Path("/Applications/Xcode-1.2.3-beta.2.app")!, version: Version(xcodeVersion: "1.2.3 Beta 2")!),
+        
+        InstalledXcode(path: Path("/Applications/Xcode-4.5.6-beta.1.app")!, version: Version(xcodeVersion: "4.5.6 Beta 1")!),
+        InstalledXcode(path: Path("/Applications/Xcode-4.5.6-beta.2.app")!, version: Version(xcodeVersion: "4.5.6 Beta 2")!),
```
senmu, line 26: Was hoping to see this scenario tested—and it is! 🙌

senmu (approved):
Code changes look great and thanks for the easy to follow testing instructions. Installing the 12.3 beta using fuzzy matching worked for me.

PR #120, opened 2020-11-23
renamed --url flag to --path for consistency

art-divin:
Small change that I've found helpful and more concise than using "--url" flag for command "install" is "--path" flag. "--url" implies that it's possible to specify a remote URL where the downloaded version would be put. But the actual implementation doesn't take this into account while relying on Path dependency (which makes total sense)

### DIFF
--- a/Sources/XcodesKit/XcodeInstaller.swift
+++ b/Sources/XcodesKit/XcodeInstaller.swift
@@ -163,7 +163,7 @@ public final class XcodeInstaller {
 
                 switch installationType {
                 case .path:
-                    // If the user provided the URL, don't try to recover and leave it up to them.
+                    // If the user provided the path, don't try to recover and leave it up to them.
                     throw error
                 default:
                     // If the XIP was just downloaded, remove it and try to recover.
--- a/Sources/XcodesKit/XcodeInstaller.swift
+++ b/Sources/XcodesKit/XcodeInstaller.swift
@@ -129,7 +129,7 @@ public final class XcodeInstaller {
     
     public enum InstallationType {
         case version(String)
-        case url(String, Path)
+        case path(String, Path)
         case latest
         case latestPrerelease
     }
@@ -162,7 +162,7 @@ public final class XcodeInstaller {
                 guard attemptNumber < 1 else { throw error }
 
                 switch installationType {
-                case .url:
+                case .path:
                     // If the user provided the URL, don't try to recover and leave it up to them.
                     throw error
                 default:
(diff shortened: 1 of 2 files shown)


Sources/xcodes/main.swift
```diff
@@ -119,7 +119,7 @@ let install = Command(usage: "install <version>",
         installation = .latest
     } else if flags.getBool(name: "latest-prerelease") == true {
         installation = .latestPrerelease
-    } else if let url = flags.getString(name: "url"), let path = Path(url) {
+    } else if let url = flags.getString(name: "path"), let path = Path(url) {
         installation = .url(versionString, path)
```
interstateone, line 123: Could we rename `XcodeInstaller.InstallationType.url` to `XcodeInstaller.InstallationType.path` to match?

Sources/xcodes/main.swift
```diff
@@ -119,7 +119,7 @@ let install = Command(usage: "install <version>",
         installation = .latest
     } else if flags.getBool(name: "latest-prerelease") == true {
         installation = .latestPrerelease
-    } else if let url = flags.getString(name: "url"), let path = Path(url) {
+    } else if let url = flags.getString(name: "path"), let path = Path(url) {
```
interstateone, line 122:
```suggestion
    } else if let pathString = flags.getString(name: "path"), let path = Path(pathString) {
```

This might be more consistent, too.

interstateone (approved):
Thanks for the PR @art-divin! This is a good change. I think I'd wanted to support remote URLs when I first wrote this, but never got around to adding support, so it must not have been that important 😄 This should be more clear for users.

There are a couple more places where the code uses "url" that I think we could change. Let me know whether you're able to update those as well, before I merge this. If not then I can take a look.

art-divin:
I have changed what you've proposed and committed just now. Please have a look! Hope there's nothing else left, but if there is - please let me know. I will update as soon as possible!

Sources/XcodesKit/XcodeInstaller.swift
```diff
@@ -162,7 +162,7 @@ public final class XcodeInstaller {
                 guard attemptNumber < 1 else { throw error }
 
                 switch installationType {
-                case .url:
+                case .path:
                     // If the user provided the URL, don't try to recover and leave it up to them.
```
interstateone, line 166:
```suggestion
                    // If the user provided the path, don't try to recover and leave it up to them.
```
One tiny last thing 😄

art-divin: Done

Issue #100, opened 2020-08-19
Support custom Xcode install location


conath:
I have limited disk space on my boot volume, so I need to download and install Xcode on an external SSD. It would be great if there was an option to pass the output folder for download and install to xcodes, preferably by using an environment variable.

interstateone:
👍 Good idea @conath! Do you prefer using an environment variable because you'd always have it set in your shell, and that way you wouldn't need to pass an argument all the time? I think we'd support both an env var and an argument, but wanted to make sure I understood first.

I'm imagining it looking something like this:
```
# command argument, which takes precedence over an environment variable if both are set
xcodes install 12.0 --destination /Volumes/Xcode

# environment variable
XCODES_DESTINATION=/Volumes/Xcode xcodes install 12.0
```

conath:
Exactly, I‘d prefer to set the environment var in my shell once (in my case it‘s /Volumes/MEDIA/Applications/), as there’s too little disk space to even download the XIP onto my boot volume.

interstateone:
> there’s too little disk space to even download the XIP onto my boot volume.

Ah, this might not solve the entire problem for you then. URLSession.downloadTask first downloads to a temporary directory outside our control which is on the "main" volume. I don't think we'd switch away from that API because it allows us to restart a failed download from where it left off which is valuable for large .xips. With this change xcodes would move the downloaded file from the temp directory to XCODES_DESTINATION on the other volume after it finished downloading and before it expanded the .xip, so I still think this would be a valuable thing to add.

conath:
You‘re right, it would still be valuable to have the option to choose where the XIP is expanded. I could probably clear enough disk space for the initial download, but definitely not for the expand/install.

art-divin:
@conath 
Unfortunately XIP command works as such it creates a cache location in /private/var/folders even if to "unxip" on an external drive completely out of built-in drive (running macOS).
I could not find a hack that would allow to completely skip the caching part.

(To be clear, xip command line utility works as:
1. expand into /private/var/folders/...
2. move expanded archive either to the location specified as an argument or into the current directory)

PR #121, opened 2020-11-23
download command to download without installing/expanding

### LINKED ISSUE #112: Allow downloading XIP without installing
There should be a `--no-install` flag that also requires a `--destination` option (see #100) (since keeping the XIP in application support and moving it to /Applications are both wrong for this use case), which downloads the XIP and then finishes without installing it. This would be useful for caching purposes.

If using aria2 to download, download the XIP directly to the destination directory instead of to application support only to move it.

art-divin:
Closes #112 

Sometimes it's useful to just download Xcode xip without expanding it.
The implemented solution shows progress as:
`xcodes download 12.0
(1/1) Downloading Xcode 12.0.0: 5 %`
where 1/6 step indicator is replaced by 1/1 depending on the used command (install vs. download)

When the download is finished, the following message appears:
`Xcode 12.0.0 has been downloaded to /Users/<#redacted#>/Library/Application Support/com.robotsandpencils.xcodes/Xcode-12.0.0.xip`

Flags and arguments are exactly as in "install" command, as such the actual code that creates "install" command is reused for both "download" and "install", and depending on the "shouldInstall" flag, one of two is created with appropriate --help configuration and other strings.

### DIFF
--- a/Sources/XcodesKit/XcodeInstaller.swift
+++ b/Sources/XcodesKit/XcodeInstaller.swift
@@ -77,7 +77,7 @@ public final class XcodeInstaller {
 
     /// A numbered step
     enum InstallationStep: CustomStringConvertible {
-        case downloading(version: String, progress: String, shouldInstall: Bool)
+        case downloading(version: String, progress: String, willInstall: Bool)
         case unarchiving
         case moving(destination: String)
         case trashingArchive(archiveName: String)
@@ -85,7 +85,12 @@ public final class XcodeInstaller {
         case finishing
 
         var description: String {
-            "(\(stepNumber)/\(stepCount)) \(message)"
+            switch self {
+            case .downloading(_, _, let willInstall) where !willInstall:
+                return "(\(stepNumber)/\(InstallationStep.downloadStepCount)) \(message)"
+            default:
+                return "(\(stepNumber)/\(InstallationStep.installStepCount)) \(message)"
+            }
         }
 
         var message: String {
@@ -116,13 +121,12 @@ public final class XcodeInstaller {
             }
         }
 
-        var stepCount: Int {
-            switch self {
-                case .downloading(_, _, let shouldInstall) where !shouldInstall:
-                    return 1
-                default:
-                    return 6
-            }
+        static var downloadStepCount: Int {
+            return 1
+        }
+        
+        static var installStepCount: Int {
+            return 6
         }
     }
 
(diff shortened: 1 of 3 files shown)


interstateone:
Thanks for this PR too! This'll be great to have, and I think `xcodes download` is more straightforward than `xcodes install --no-install` as I'd described in the issue.

I'd prefer to have this `download` command save the XIP to a path that the user provides, such as with a `--destination` flag, so that the application support directory isn't a place the user needs to know about. This is related to #100, and I have some work in progress to implement this locally that I'll try to finish up soon. I don't think that work necessarily needs to prevent merging this PR, so I may make a few changes after merging this and before releasing a new version.

Just wanted to let you know that I've seen and appreciate this PR, and it may take a bit more time to review and merge it. Thanks @art-divin.

art-divin:
> xcodes download is more straightforward than xcodes install --no-install
Could not agree more! I already use this option a lot. 

> I may make a few changes after merging this and before releasing a new version
Sounds just great! You can open more issues if you have something in mind you can delegate, they may be related to what I am up to.

I am also implementing option to allow `xcodes download 12.0 --remove` to support clean up, as well as `xcodes download --clear-cache` or similar naming - to remove all of the downloaded xips that are within xcodes' app container.

Later (this week will be delivered) I will be implementing commands such as `xcodes unxip` (already in progress), `xcodes undmg` (currently dmg is unsupported), `xcodes finder 12.1` (open directory containing the downloaded xip / unxipped .app of the specified version) and maybe some more. FYI I will be doing this myself anyways, they will be used in my current project. I will be creating PRs so you are free to merge them whenever you'd like!

I appreciate your feedback, though, don't waste time thanking me, we all are beneficiaries anyways :smile:

interstateone:
Hmm, those commands don't sound like something I'd considered before. Could you tell me more about the problem you're solving and how you're planning on using them?

art-divin:
> Could you tell me more about the problem you're solving and how you're planning on using them?
Ha thanks for your interest! 😅
Unfortunately *this* project needs to be properly finished before I can tell, but you can be sure there’s going to be your mention in 3rd-party licenses section!

art-divin:
I've discovered that `--path` is not taken into account when using the new `download` command. The reason is that in the old version, `install` command was using the provided `--path` as an input for the actual downloaded `xip` file. However, w/ the case of `download` command, that argument needs to be taken as a destination of the downloaded file to be moved in both URLSession and aria2 variants.
I've fixed this in my fork, but I need to cherry-pick changes to this branch before I'll be able to commit. So, fix is coming!

art-divin:
OK, added support for --path flag, so
`xcodes download --path ~/Downloads/xcodes/something.xip` would move file (downloaded by NSURLSession) into the --path directory, 
the same for --aria2 option, except that cache for --aria2 would also be placed inside of that destination directory

ccarpita:
pr-great-interest

interstateone:
Thanks for your patience @art-divin, I'm taking another look at this now. I'm going to rebase it on master and make a few changes.

I'm going to rename the `path` argument to `directory`, which will match the argument name in my implementation of #100, and also avoids confusion of a path to the _directory_ and a full path to the _.xip_.

I'm also going to set a default value for `directory` of `~/Downloads` and move the .xip there, as I don't think the application support directory should be exposed to the user.

---

> Also, there's an issue w/ URLSession option. The cache can become both invalid and large in size if there'd be multiple attempts of download w/ NSURLSessionDownloadTask. There's no way to specify URL for the cache files, they are put in /private/var/folders/... (Which is eventually gets cleaned by the system under some specific conditions, but there's no manual control over it as far as I am concerned.). Compared to aria2 option, which creates cache file in the same dir as where the final destination is.

I think this behaviour is fine for xcodes' purposes, unless there's a particular reason you think the in-progress download data should be handled differently?

---

Thanks for the contribution!

interstateone (approved):

PR #125, opened 2020-12-30
fix(XcodeInstaller): Infinite auth fail loop

ccarpita:
If the ENV password is incorrect, the `loginWithNeeded` method will fail in a recurring loop. This change will ensure that STDIN input is required if authentication fails the first time. Also, the current username is printed out on the password input line for improved clarity.

### DIFF
--- a/Sources/XcodesKit/XcodeInstaller.swift
+++ b/Sources/XcodesKit/XcodeInstaller.swift
@@ -282,16 +282,32 @@ public final class XcodeInstaller {
         }
     }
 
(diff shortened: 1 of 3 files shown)


interstateone (approved):
Thanks for identifying this bug and for the contribution @ccarpita! I like the idea of including the Apple ID when prompting for a password, too. I made a small change to only do this when the user didn't just enter their Apple ID, and I added a test case that covers this bug.

ccarpita:
Thanks for making the changes and merging!

PR #126, opened 2020-12-31
Add directory argument

### LINKED ISSUE #100: Support custom Xcode install location

interstateone:
This adds a directory argument at the top level (so it affects all commands) which determines the directory that xcodes installs in to. It defaults to /Applications, which is the current behaviour. xcodes will also check for a value in the XCODES_DIRECTORY environment variable. It applies to installation but also the other commands, so that e.g. selecting or uninstalling a version looks in the desired directory.

The motivation for all of this is to support users that may not want to store Xcode versions in /Applications, for example if they don't have enough space on their boot volume.

This also aligns with the addition of the directory argument as part of #121.

## Testing

```
cd xcodes
git checkout -b interstateone-xcodes-destination master
git pull https://github.com/interstateone/xcodes.git xcodes-destination
```

then run commands with either the directory argument or XCODES_DIRECTORY environment variable set. If it's set to a directory that doesn't exist, or to a path that isn't a directory, it should print an error.

Expected results:

```
❯ swift run xcodes list --directory /Users/brandon/Xcodes
Directory argument must be a directory, but was provided /Users/brandon/Xcodes.

# make sure the directory exists
❯ mkdir /Users/brandon/Xcodes

# it doesn't have any Xcodes in it yet
❯ swift run xcodes installed --directory /Users/Brandon/Xcodes

# download command still works like before
❯ swift run xcodes download 12.3 --directory /Users/brandon/Xcodes
(1/1) Downloading Xcode 12.3.0: 99%

Xcode 12.3.0 has been downloaded to /Users/brandon/Xcodes/Xcode-12.3.0.xip

❯ ls /Users/brandon/Xcodes | grep Xcode                               
-rw-r--r--   1 brandon  staff    11G  2 Jan 11:19 Xcode-12.3.0.xip

# install works
❯ XCODES_DIRECTORY=/Users/brandon/Xcodes swift run xcodes install 12.2                                  
(1/6) Downloading Xcode 12.2.0: 99%
(3/6) Moving Xcode to /Users/brandon/Xcodes/Xcode-12.2.0.app
(4/6) Moving Xcode archive Xcode-12.2.0.xip to the Trash
(5/6) Checking security assessment and code signing
(6/6) Finishing installation
xcodes requires superuser privileges in order to finish installation.
macOS User Password: 

Xcode 12.2.0 has been installed to /Users/brandon/Xcodes/Xcode-12.2.0.app

❯ ls /Users/brandon/Xcodes | grep Xcode
drwxr-xr-x   3 brandon  staff    96B 23 Oct 20:24 Xcode-12.2.0.app
-rw-r--r--   1 brandon  staff    11G  2 Jan 11:19 Xcode-12.3.0.xip
```

---

Closes #100

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,9 @@ Install a specific version of Xcode using a command like one of these:
 xcodes install 10.2.1
 xcodes install 11 Beta 7
 xcodes install 11.2 GM seed
-xcodes install --latest
+xcodes install 9.0 --path ~/Archive/Xcode_9.xip
+xcodes install --latest-prerelease
+xcodes install --latest --directory "/Volumes/Bag Of Holding/"
 ```
 
 You'll then be prompted to enter your Apple ID username and password. You can also provide these with the `XCODES_USERNAME` and `XCODES_PASSWORD` environment variables.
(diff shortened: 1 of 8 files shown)


csykes (approved):
Verified locally using ENV variable + command line option

PR #129, opened 2021-01-24
Add support for and default to Xcode Releases data source

### LINKED ISSUE #123: Release candidate prerelease identifiers aren't being handled correctly
E.g.
```
12.2 Release Candidate
12.2 (Installed, Selected)
12.3 Beta (12C5020f) (Installed)
12.3 Rc (12C33)
12.3 Release Candidate
```

The last two versions should be treated as the same. Expanding `RC` to `Release Candidate` when parsing might be sufficient.

interstateone:
This adds support for [Xcode Releases](https://xcodereleases.com) as a data source for available Xcode versions, and makes it the default. It's still possible to use the Apple data source by providing the `--data-source=apple` option. The motivation for this is that while the data is unofficial it is more complete and reliable than scraping the Apple Developer website. This code is all taken from (with some small changes) [Xcodes.app](https://github.com/RobotsAndPencils/XcodesApp).

Two other changes come along with this:

- Previously build identifiers (for example "12C33" for the final 12.3 release) were only shown for non-final release versions. I think there's value in showing this for all versions since sometimes you need to figure out exactly what versions are installed or available, for example when Apple releases a final version that was the same build as a release candidate. This is now done when versions are printed in a list (list, installed, select).
- Version comparison has been simplified in most cases. It's still a little fuzzy when parsing version strings from the user, which is expected.

## Testing

`swift run xcodes update` first to make sure you have Xcode Releases data, and then use other commands and make sure they still work the way you expect.

You can do the same after running `swift run xcodes update --data-source=apple` to make sure the Apple one still works, except for the issue noted in #123.

---

I'm going to say that this fixes #123 for now. We could invest more time into making the Apple data source work perfectly but part of the motivation for adding Xcode Releases support is to not have to do that. If another contributor wishes to do so we'd welcome that change.

### DIFF
--- a/Sources/XcodesKit/Version+.swift
+++ b/Sources/XcodesKit/Version+.swift
@@ -1,47 +1,24 @@
 import Version
 
 public extension Version {
-    func isEqualWithoutBuildMetadataIdentifiers(to other: Version) -> Bool {
-        return major == other.major && 
-               minor == other.minor &&
-               patch == other.patch &&
-               prereleaseIdentifiers == other.prereleaseIdentifiers
-    }
-
-    func isEqualWithoutAllIdentifiers(to other: Version) -> Bool {
-        return major == other.major &&
-               minor == other.minor &&
-               patch == other.patch
-    }
-
-    /// If release versions, don't compare build metadata because that's not provided in the /downloads/more list
-    /// if beta versions, compare build metadata because it's available in versions.plist
-    func isEquivalentForDeterminingIfInstalled(toInstalled installed: Version) -> Bool {
-        let isBeta = !prereleaseIdentifiers.isEmpty
-        let otherIsBeta = !installed.prereleaseIdentifiers.isEmpty
-
-        if isBeta && otherIsBeta {
-            if buildMetadataIdentifiers.isEmpty {
-                return major == installed.major &&
-                       minor == installed.minor &&
-                       patch == installed.patch &&
-                       prereleaseIdentifiers == installed.prereleaseIdentifiers
-            }
-            else {
-                return major == installed.major &&
-                       minor == installed.minor &&
-                       patch == installed.patch &&
-                       prereleaseIdentifiers == installed.prereleaseIdentifiers &&
-                       buildMetadataIdentifiers.map { $0.lowercased() } == installed.buildMetadataIdentifiers.map { $0.lowercased() }
-            }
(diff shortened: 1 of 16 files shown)


Sources/XcodesKit/Version+.swift
```diff
@@ -1,47 +1,24 @@
...
+    /// 
+    /// We need a way to determine if two Xcode versions are the same without always having full information, and supporting different data sources.
+    /// For example, the Apple data source often doesn't have build metadata identifiers.  
+    func isEquivalent(to other: Version) -> Bool {
+        // If we don't have build metadata identifiers for both Versions, compare major, minor, patch and prerelease identifiers.
+        if buildMetadataIdentifiers.isEmpty || other.buildMetadataIdentifiers.isEmpty {
+            return major == other.major &&
+                   minor == other.minor &&
+                   patch == other.patch &&
+                   prereleaseIdentifiers.map { $0.lowercased() } == other.prereleaseIdentifiers.map { $0.lowercased() }
```
senmu, line 14: The first 3 checks have been pulled up into `isEqualWithoutAllIdentifiers(to other: Version) -> Bool`. Could that be incorporated here to avoid repeating?
```suggestion
            return isEqualWithoutAllIdentifiers(to: other) &&
                   prereleaseIdentifiers.map { $0.lowercased() } == other.prereleaseIdentifiers.map { $0.lowercased() }
```

Sources/XcodesKit/Version+.swift
```diff
@@ -1,47 +1,24 @@
...
+        // If we have build metadata identifiers for both, we can ignore the prerelease identifiers.
+        } else {
+            return major == other.major &&
+                   minor == other.minor &&
+                   patch == other.patch && 
+                   buildMetadataIdentifiers.map { $0.lowercased() } == other.buildMetadataIdentifiers.map { $0.lowercased() }
```
senmu, line 20: Same comment here about utilizing `isEqualWithoutAllIdentifiers(to other: Version) -> Bool`. Can it also be used here to avoid repeating?
```suggestion
            return isEqualWithoutAllIdentifiers(to: other) && 
                   buildMetadataIdentifiers.map { $0.lowercased() } == other.buildMetadataIdentifiers.map { $0.lowercased() }
```

senmu (approved):
Works great for me. Just had a minor suggestion to avoid repetition

PR #130, opened 2021-01-24
Add support for coloured output

interstateone:
This adds Rainbow as a dependency for coloured output.

I find this helps in two situations:

- When looking at a long list of information, like output from `xcodes list`, it's easier to visually scan for lines that are coloured in some way. In this specific case, Installed is printed in blue and Selected is printed in green.
- Warnings, errors and success during long-running processes, like downloading and installing, are easier to identify at a glance when it's coloured yellow, red, or green, respectively.

|list|installed|select|error|no color|
|-----|-----|-----|-----|-----|
| Screen Shot 2021-02-02 at 7 31 18 PM | Screen Shot 2021-02-02 at 7 31 35 PM | Screen Shot 2021-02-02 at 7 32 09 PM | Screen Shot 2021-02-02 at 7 32 25 PM | Screen Shot 2021-02-02 at 7 35 19 PM |

I've added a global `--color/no-color` flag that defaults to true/enabled. If xcodes is running in an interactive terminal then it will use color by default. It will disable coloured output if the user provides the `--no-color` flag, if its not running in an interactive terminal, if the NO_COLOR environment variable is set, or if the TERM environment variable is set to "dumb". This is following the suggestions in [Command Line Interface Guidelines](https://clig.dev/#output)

---

One thing that would be good to do next is expand on Logging.log with different logging levels, like info, debug, warn and error. That would enable two things:

- A global verbose flag to enable debug-level logs for debugging and testing purposes
- Colors could be set for the warn and error levels in one place, instead of at the call site. We could also send errors to STDERR, which would be more appropriate.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -56,6 +56,15 @@
         }
       },
       {
+        "package": "Rainbow",
+        "repositoryURL": "https://github.com/onevcat/Rainbow.git",
+        "state": {
+          "branch": null,
+          "revision": "626c3d4b6b55354b4af3aa309f998fae9b31a3d9",
+          "version": "3.2.0"
+        }
+      },
+      {
         "package": "swift-argument-parser",
         "repositoryURL": "https://github.com/apple/swift-argument-parser",
         "state": {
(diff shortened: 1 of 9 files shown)


senmu (approved):
Works great! :+1:

Issue #23, opened 2019-03-20
Print the path of installed versions when running the installed command


interstateone:
The current behaviour prints the versions that are installed:
```
$ xcodes installed
12.2
12.3 (Selected)
12.4 Release Candidate (12D4e)
```

It may be useful to also print the path they're installed at, perhaps like:

```
$ xcodes installed
12.2	/Applications/Xcode-12.2.0.app
12.3 (Selected)	/Applications/Xcode-12.3.0.app
12.4 Release Candidate (12D4e)	/Applications/Xcode-12.4.0-release.candidate.app
```

There should be a tab character in between the current output and the path so it's easier to split the lines into their components. For example, you could get a specific path with:

```sh
$ xcodes installed | \ # get all installed versions and their paths
    grep 12.2 | \ # get the line for version 12.2
    cut -d$'\t' -f2 # split by tab character, return second field
/Applications/Xcode-12.2.0.app
```

interstateone:
I can’t remember why I wanted this. Maybe to grep and pipe to xcode-select? There’s a select command in xcodes now though.

interstateone:
I've added some more information to the issue and will re-open it, as it might be useful for some use cases like getting a path to set the DEVELOPER_DIR environment variable in scripts.

PR #133, opened 2021-01-28
Prints Xcode path for installed versions

NinjaLikesCheez:
This change addresses #23 by adding the path of installed Xcode versions to the output of the `installed` command.


interstateone (approved):
Awesome, thanks!

Issue #136, opened 2021-02-03
Add signout command


interstateone:
Add a signout command so that users can more easily sign out or switch accounts. Currently the latter requires using the `XCODES_USERNAME` environment variable which is not very discoverable.

- [x] Remove all Keychain items
- [x] Remove cookies in the shared URLSession
- [x] Set `defaultUsername` in Configuration to nil
- [x] If #130 has merged, use coloured output for success, warning or failure messages

evgeniyd:
Hey. Just wonder, is there any plan to release the version with this change? 
Building from source indeed works for me, but seeing the feature as an official release would streamline the process.

Thanks.

PR #140, opened 2021-03-17
Add signout command

senmu:
Addresses enhancement #136

## What changed
A new `signout` command has been added to:
* Remove all Keychain items
* Remove cookies in the shared URLSession
* Set `defaultUsername` in Configuration to nil

xcodes-signout

### DIFF
--- a/Sources/AppleAPI/Client.swift
+++ b/Sources/AppleAPI/Client.swift
@@ -29,7 +29,7 @@ public class Client {
             case .noTrustedPhoneNumbers:
                 return "Your account doesn't have any trusted phone numbers, but they're required for two-factor authentication. See https://support.apple.com/en-ca/HT204915."
             case .notAuthenticated:
-                return "No authenticated session found"
+                return "You are already signed out"
             default:
                 return String(describing: self)
             }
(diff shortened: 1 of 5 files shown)


Sources/AppleAPI/Client.swift
```diff
@@ -27,6 +28,8 @@ public class Client {
                 return "Not a valid phone number index. Expecting a whole number between \(min)-\(max), but was given \(given ?? "nothing")."
             case .noTrustedPhoneNumbers:
                 return "Your account doesn't have any trusted phone numbers, but they're required for two-factor authentication. See https://support.apple.com/en-ca/HT204915."
+            case .notAuthenticated:
+                return "No authenticated session found"
```
MattKiazyk, line 32: image
Do you think the red is a little harsh for the error? It really just is an info statement. Perhaps the description here would be better a little less error like. `You are already signed out` ?

MattKiazyk (approved):
All good Sam - just one picky comment out the language.. take it or leave it. Nothing to die on.

Sources/AppleAPI/Client.swift
senmu: I like both suggestions 👍 
Implemented it in d49c722

PR #143, opened 2021-04-22
Forward legible errors

ffittschen:
I'm currently trying to uninstall Xcode 12.5 Beta 3 using `xcodes`, but it fails with the following error:

```bash
$ xcodes uninstall 12.5 Beta 3
The operation couldn’t be completed. (ExitCode(rawValue: 1))
```

Upon investigating the cause, I discovered that the error is not correctly forwarded, which is why I can't see the _real_ error message. This PR fixes that and forwards the catched errors.

### DIFF
--- a/Sources/xcodes/main.swift
+++ b/Sources/xcodes/main.swift
@@ -300,7 +300,7 @@ struct Xcodes: ParsableCommand {
                 }
             }
             .done { List.exit() }
-            .catch { error in List.exit(withLegibleError: ExitCode.failure) }
+            .catch { error in List.exit(withLegibleError: error) }
             
             RunLoop.current.run()
         }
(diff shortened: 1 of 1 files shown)


MattKiazyk:
Thanks for the PR @ffittschen - I'm curious what the result was on your machine with that change?

ffittschen:
I didn't see the result, but I figured out the issue:

We are using `xcodes` within ansible playbooks to manage our Xcode installations on the CI machines. As the `xcodes install` command asks interactively asks for a password, we run the command using `become: true` to avoid this. As it turns out, subsequently calling `xcodes uninstall` without `become: true` results in the error I posted above. But when I add `become: true` to the uninstall call, the Xcode.app is actually moved to `/private/var/root/.Trash/` instead of `~/.Trash/`. As a workaround I added an "Empty trash" task after the uninstall task, but it's not perfect 😄

Do you happen to have any suggestions on how to run `xcodes install / uninstall` without `sudo` or `become: true` in a non-interactive way?

MattKiazyk (approved):

MattKiazyk:
@ffittschen I'm going to merge this in and release it. If you want - can you open up an issue to talk about the other issue you've brought up. 

Thanks for the PR - sorry it took so long to get through

PR #172, opened 2021-12-16
Setting universal build

### LINKED ISSUE #170: Support arm64 (M1) builds
When trying to install the latest version of xcodes through Homebrew the binary it installs is only for arm64.

This means xcodes no longer runs on Intel based Macs.

This seems to be an isolated issue to v0.18.0.

cc/ @MattKiazyk

fishcharlie:
Close #170

@MattKiazyk Might want to test this before merging. It looks like **maybe** the output executable might be in a different location. But since I'm not completely familiar with the build/distribution system of this project, I'm not sure that matters.

I was able to verify that the outputted executable from this PR does produce a universal binary tho.

Please let me know if any adjustments need to be made.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,8 @@ xcodes: $(SOURCES)
 		-Xswiftc -Onone \
 		--disable-sandbox \
 		--build-path "$(BUILDDIR)" \
+		--arch arm64 \
+		--arch x86_64 \
 
 .PHONY: sign
 sign: xcodes


MattKiazyk (approved):

MattKiazyk:
@fishcharlie 🙏 🙏 

As far as I can tell this is working as advertised! We'll find out for sure next release, but I'll for sure double check on an M1 (now that I have one) and the old intel one

jpsim:
Worked great for me on my M1, installed via the `xcodes.zip` archive from https://github.com/RobotsAndPencils/xcodes/releases/tag/0.20.0

MattKiazyk:
👍 - it doubled the zip from 1.8 to 3.6 MB 😄

jpsim:
By law, no one is allowed to complain about 3.6MB for a tool used to download 10+ GB files from the internet 😆

PR #179, opened 2022-01-23
Moved XIP expansion to a temporal directory

### LINKED ISSUE #178: Expand .xip on the destination volume
After PR #126, `xcode install` supports a custom path for Xcode's installation. Unfortunately, this path isn't used for the .xip expansion, using instead the same directory where it’s downloaded (`~/Library/Application Support/com.robotsandpencils.xcodes/`)[^1]. If this custom installation path is to a different volume than the .xip, this means that:

- Xcode.app must be first expanded and then moved to another volume, a relatively slow operation given its massive size.
- This installation requires double the free space than on a same-volume installation, because Xcode is first copied to the destination volume and only when it has finished it’s deleted from the original volume.

So I think it makes sense to expand the .xip directly on the final volume where Xcode will be installed (probably on the `/.TemporaryItems` folder).

Would you accept a pull request for this? Thank you!

[^1]: I assume this is done to avoid having a partial Xcode installation on the destination folder, which sounds reasonable to me.

juanjonol:
Instead of expanding XIPs on its current directory, a temporal directory is used. The main advantage of this approach is that, because the temporal directory is created on the same volume where Xcode will be installed, the installation is vastly improved if the XIP is on a different volume.

- It also makes more conceptual sense to keep a temporal file on the OS temporal folders.
- https://nshipster.com/temporary-files/
- The implementation of the `temporalDirectory` function and variable are based on the similar `trashItem`, but without the `@discardableResult` annotation (I cannot think of any case where this URL could to be ignored).

This closes #178.

___

I've tested this on September 21st, to get some hard data about the improvement with this PR.

Installing Xcode 14.1 beta 2 (14B5024i) on a different volume that the one where the .xip is located, the `(3/6) Moving Xcode to /Volumes/<path>/Xcode-14-1-0-Beta.2.app` phase with the old implementation (using the `--xip-expand-inplace` flag) takes 4.78 minutes on a 2021 MacBook Pro 14". Without `--xip-expand-inplace`, this is virtually instantaneous (0.0009 seconds).

For comparison, the improvement I get using `--experimental-unxip` over regular unxip is 1.33 minutes.


juanjonol:
A few comments I didn't add to the commit:

- This always changes the expansion directory, not just when the destination volume is different that the one where the XIP is. This means the XIP will never be expanded on `~/Library/Application Support/com.robotsandpencils.xcodes/` again.[^1] I did this for consistency and because I don't think there is anything useful that could be done with a partially-expanded Xcode version. This also allows macOS to automatically delete the expanded data if for some reason `xcodes` fails to move it.
- @MattKiazyk suggested [a change similar to other one on Xcodes.app](https://github.com/RobotsAndPencils/XcodesApp/pull/155/files), but I'm not sure if it makes sense here, for the same reasons as the previous point.

[^1]: Theoretically, `url(for:in:appropriateFor:create:)` can fail, and in that case the original code will be used, but I haven't found any case where it actually fails.

Sources/XcodesKit/XcodeInstaller.swift
```diff
@@ -708,9 +708,10 @@ public final class XcodeInstaller {
     }
 
     func unarchiveAndMoveXIP(at source: URL, to destination: URL) -> Promise<URL> {
+        let xcodeExpansionDirectory = (try? Current.files.temporalDirectory(for: source)) ?? source.deletingLastPathComponent()
```
MattKiazyk, line 711: #picky Can we move the `(try? Current.files.temporalDirectory(for: source)) ?? source.deletingLastPathComponent()` into it's own function inside of `Files` so that we can just call `try? Current.files.xcodeExpansionDirectory`

MattKiazyk (changes requested):

MattKiazyk:
@juanjonol Thanks for the PR! It is appreciated! I love what you did here. I have one picky comment plus another ask.

Can we add a flag onto the cli so that the user could (if they wanted) to NOT expand to the temporary directory and instead to where it was downloaded? I agree temporary is better, but just wanted to give that option to the user if they wanted to revert back to the old behaviour

Thanks!

Sources/XcodesKit/XcodeInstaller.swift
juanjonol: I tried to solve this on https://github.com/RobotsAndPencils/xcodes/pull/179/commits/f1163aaf45111dcca1bc9a664fd8f49e7240ee3b. Please take a look! And don't worry about being picky 😊

juanjonol:
@MattKiazyk I added a `expand-xip-inplace` flag to keep the old behaviour (simply using the directory where the .xip is). If you need anything else or you don't like how it's named/implemented, don't hesitate to tell me (I'm picky too!).

Also, I though about adding tests, but I don't know how to test this. It's a small change so maybe it isn't needed, but if you have any idea about it, we can look into it.

Sources/XcodesKit/XcodeInstaller.swift
juanjonol: I mark this as resolved to remove some noise from this Pull Request, please fill free to reopen it if you have some comments!

juanjonol:
Hi @MattKiazyk, this PR has been stale for a few months. Could you please review it? I see there're conflicts now, but there's no point fixing them if the PR won't be merged...

juanjonol:
I've updated this PR with the latest changes on main. All test are green and it can be merged without conflicts.

@MattKiazyk could you please take a look at this and tell me if there's something preventing its merge? This still says that there're changes requested, but I made those changes months ago...

Issue #142, opened 2021-03-30
Don't exit(1) on certain errors, when they're not true failures. 


tahirmt:
Currently if I run `xcodes install --latest` the command fails if Xcode is already installed. Since it is already installed, do not fail the command

tahirmt:
@MattKiazyk Yes I see this but it exists after and doesn't continue. I have `xcodes` as part of a script that has other steps after this but xcodes exits my script because Xcode is already installed

MattKiazyk:
Ah I see what you're meaning. Thanks for that. For sure we shouldn't be exiting hard in this scenario.

julianxhokaxhiu:
I recently crossed exactly this issue and I was wondering if some progress have been done related.

In my own case this is the error:
```sh
$ xcodes install 12.4
12.4 is already installed at /Applications/Xcode-12.4.0.app
make: *** [Makefile:50: install_xcode_versions] Error 1
```

I could bypass this by adding `|| true` at the moment, but I would be nice if you could maybe not exit with an error code > 0.

Thank you in advance!

PR #180, opened 2022-01-31
Do not exit(1) if xcode is already installed

tahirmt:
When running `xcodes install x.x.x` if the Xcode version is already installed, do not exit(1). Instead, exit with a success.

Addresses https://github.com/RobotsAndPencils/xcodes/issues/142

Example:

Screen Shot 2022-01-30 at 11 01 41 PM


MattKiazyk:
Thanks @tahirmt this looks really good. Thanks for adding it in

MattKiazyk (approved):

julianxhokaxhiu:
Thank you!

PR #181, opened 2022-01-31
Exit immediately if trying to select currently selected xcode

tahirmt:
If the currently requested xcode is already selected, there is no need to prompt the user to give sudo permissions and re-select the path. We can just avoid the step and exit early.

Closes https://github.com/RobotsAndPencils/xcodes/issues/102

PR #182, opened 2022-02-04
Add select option to install

### LINKED ISSUE #161: [Feature Request] xcodes install *** --select
Whenever I install a new version of xcode, I always want to select it at the same time. It'd be great if we could combine the `select` with the install.

```bash
> xcodes install 13.1 --select
# or
> xcodes install --latest --select
```

tahirmt:
Adds options `--select` and `--update` to `install` command.

Details
---
* Uses #181 as base because without that the `--select` option will always ask user for password even if the Xcode version is already selected
* I'm not that fluent in `PromiseKit` so happy to make any improvements
* Creating this as a draft until we can get #181 merged
* Closes #161

### DIFF
--- a/Sources/XcodesKit/XcodeInstaller.swift
+++ b/Sources/XcodesKit/XcodeInstaller.swift
@@ -177,8 +177,8 @@ public final class XcodeInstaller {
     }
 
     /// Perform the install but don't exit out but return the installed xcode version as output instead
-    public func installWithoutLogging(_ installationType: InstallationType, dataSource: DataSource, downloader: Downloader, destination: Path, experimentalUnxip: Bool) -> Promise<InstalledXcode> {
-        self.install(installationType, dataSource: dataSource, downloader: downloader, destination: destination, attemptNumber: 0, experimentalUnxip: experimentalUnxip)
+    public func installWithoutLogging(_ installationType: InstallationType, dataSource: DataSource, downloader: Downloader, destination: Path, experimentalUnxip: Bool = false, emptyTrash: Bool, noSuperuser: Bool) -> Promise<InstalledXcode> {
+        self.install(installationType, dataSource: dataSource, downloader: downloader, destination: destination, attemptNumber: 0, experimentalUnxip: experimentalUnxip, emptyTrash: emptyTrash, noSuperuser: noSuperuser)
     }
     
     private func install(_ installationType: InstallationType, dataSource: DataSource, downloader: Downloader, destination: Path, attemptNumber: Int, experimentalUnxip: Bool, emptyTrash: Bool, noSuperuser: Bool) -> Promise<InstalledXcode> {
--- a/Sources/XcodesKit/XcodeInstaller.swift
+++ b/Sources/XcodesKit/XcodeInstaller.swift
@@ -166,17 +166,12 @@ public final class XcodeInstaller {
         case aria2(Path)
     }
 
-    public func install(_ installationType: InstallationType, dataSource: DataSource, downloader: Downloader, destination: Path, experimentalUnxip: Bool = false, emptyTrash: Bool, noSuperuser: Bool, exitAfterInstall: Bool = true) -> Promise<InstalledXcode> {
+    public func install(_ installationType: InstallationType, dataSource: DataSource, downloader: Downloader, destination: Path, experimentalUnxip: Bool = false, emptyTrash: Bool, noSuperuser: Bool) -> Promise<InstalledXcode> {
         return firstly { () -> Promise<InstalledXcode> in
             return self.install(installationType, dataSource: dataSource, downloader: downloader, destination: destination, attemptNumber: 0, experimentalUnxip: experimentalUnxip, emptyTrash: emptyTrash, noSuperuser: noSuperuser)
         }
         .map { xcode in
             Current.logging.log("\nXcode \(xcode.version.descriptionWithoutBuildMetadata) has been installed to \(xcode.path.string)".green)
-
-            if exitAfterInstall {
-                Current.shell.exit(0)
-            }
-
             return xcode
         }
     }
--- a/Sources/XcodesKit/XcodeInstaller.swift
+++ b/Sources/XcodesKit/XcodeInstaller.swift
@@ -166,19 +166,19 @@ public final class XcodeInstaller {
         case aria2(Path)
     }
 
-    public func install(_ installationType: InstallationType, dataSource: DataSource, downloader: Downloader, destination: Path, experimentalUnxip: Bool = false, emptyTrash: Bool, noSuperuser: Bool) -> Promise<Void> {
+    public func install(_ installationType: InstallationType, dataSource: DataSource, downloader: Downloader, destination: Path, experimentalUnxip: Bool = false, emptyTrash: Bool, noSuperuser: Bool, exitAfterInstall: Bool = true) -> Promise<InstalledXcode> {
         return firstly { () -> Promise<InstalledXcode> in
             return self.install(installationType, dataSource: dataSource, downloader: downloader, destination: destination, attemptNumber: 0, experimentalUnxip: experimentalUnxip, emptyTrash: emptyTrash, noSuperuser: noSuperuser)
         }
-        .done { xcode in
+        .map { xcode in
             Current.logging.log("\nXcode \(xcode.version.descriptionWithoutBuildMetadata) has been installed to \(xcode.path.string)".green)
-            Current.shell.exit(0)
-        }
-    }
 
-    /// Perform the install but don't exit out but return the installed xcode version as output instead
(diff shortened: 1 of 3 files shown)


Sources/XcodesKit/XcodeInstaller.swift
```diff
@@ -164,6 +164,11 @@ public final class XcodeInstaller {
             Current.shell.exit(0)
         }
     }
+
+    /// Perform the install but don't exit out but return the installed xcode version as output instead
+    public func installWithoutLogging(_ installationType: InstallationType, dataSource: DataSource, downloader: Downloader, destination: Path) -> Promise<InstalledXcode> {
```
Kingkhalifa195, line 169: I keep having set back anytime I installed it

Kingkhalifa195 (commented):
Will try to check this out

Sources/XcodesKit/XcodeInstaller.swift
tahirmt: Could you please elaborate?

rogerluan:
@tahirmt thank you for this PR! This makes total sense to me 💯 

@MattKiazyk this would also be aligned with https://github.com/xcpretty/xcode-install/issues/467, since xcode-install's default behavior automatically selects the Xcode version that was just installed, and it would be a huge improvement over xcode-install and xcodes if we could also optionally run `update` automatically right before installing (otherwise Xcode versions that exist might not be found). I had to work-around this behavior implementing a "pre-install update" behavior [here](https://github.com/fastlane/fastlane/pull/20672/files#diff-6497c81804abac26be66c3463195f40d2f252bddd182438e5df8d01ec266f901R14-R21) but it'd be great if it was built-in into xcodes ☺️

I'll review this PR to help move it forward 💪

Sources/xcodes/main.swift
```diff
@@ -220,17 +232,60 @@ struct Xcodes: ParsableCommand {
...
+                        Install.processDownloadOrInstall(error: error)
+                    }
+            } else {
+                // check if the version is already installed and try to select it
+                firstly { () -> Promise<Void> in
+                    if case .version(let version) = installation {
+                        return selectXcode(shouldPrint: print, pathOrVersion: version, directory: destination, fallbackToInteractive: false)
+                    } else {
+                        return Promise { _ in
+                            throw InstallError.notAlreadyInstalled
```
MattKiazyk, line 250: I'm not a huge fan of throwing just to get to a catch. Can you refactor this a bit so that it's a bit simpler around the different options. That may involved some new functions to make things a bit cleaner. 

Let me know if you need some help around this. Thanks

MattKiazyk (changes requested):

Sources/xcodes/main.swift
tahirmt: I did some refactor. Let me know what you think.

tahirmt: I think unit testing is where I'm kind of lost and will require some help

Sources/xcodes/main.swift
```diff
@@ -217,69 +229,103 @@ struct Xcodes: ParsableCommand {
...
             let destination = getDirectory(possibleDirectory: directory)
-            
-            installer.install(installation, dataSource: globalDataSource.dataSource, downloader: downloader, destination: destination, experimentalUnxip: experimentalUnxip, emptyTrash: emptyTrash, noSuperuser: noSuperuser)
-                .done { Install.exit() }
-                .catch { error in
-                    Install.processDownloadOrInstall(error: error)
+
+            if select, case .version(let version) = installation {
+                firstly {
+                    selectXcode(shouldPrint: print, pathOrVersion: version, directory: destination, fallbackToInteractive: false)
```
MattKiazyk, line 244: I think this is incorrect - we should install first, then select after.

Sources/xcodes/main.swift
```diff
@@ -217,69 +229,103 @@ struct Xcodes: ParsableCommand {
...
+                        }
+                } else {
+                    return installer.install(installation, dataSource: globalDataSource.dataSource, downloader: downloader, destination: destination, experimentalUnxip: experimentalUnxip, emptyTrash: emptyTrash, noSuperuser: noSuperuser, exitAfterInstall: !select)
+                }
+            }
+            .then { xcode -> Promise<Void> in
+                Current.logging.log("\nXcode \(xcode.version.descriptionWithoutBuildMetadata) has been installed to \(xcode.path.string)".green)
+
+                // Install was successful, now select it
+                return selectXcode(shouldPrint: print, pathOrVersion: xcode.path.string, directory: destination, fallbackToInteractive: false)
```
MattKiazyk, line 275: The `selectXcode` here isn't looking at the `select` option passed in.

MattKiazyk (changes requested):
Thanks for the PR!

There's some coding issues around how the `select` parameter gets passed through. Made some comments

Sources/xcodes/main.swift
tahirmt: This was done to only select if it is already installed. If the version isn't installed this will throw and then we can install it.

rogerluan:
@tahirmt since https://github.com/RobotsAndPencils/xcodes/pull/181 got merged, would you mind rebasing this branch or merging the main branch into this one? 🙇

Sources/xcodes/main.swift
tahirmt: Actually, this line will not get executed if select is false because install would exit early from the step before. That's why I don't have a check here.

tahirmt: Unless I missed something in how Install.exit works.

tahirmt: Yep I tested it out

```
➜  xcodes git:(add-select-option-to-install) swift run xcodes install 14.0.1 --experimental-unxip
Building for debugging...
Build complete! (0.08s)

Downloading with aria2
(1/6) Downloading Xcode 14.0.1+14A400: 99%
Using experimental unxip. If you encounter any issues, remove the flag and try again
(3/6) Moving Xcode to /Applications/Xcode-14.0.1.app
(4/6) Moving Xcode archive Xcode-14.0.1+14A400.xip to the Trash
(5/6) Checking security assessment and code signing
(6/6) Finishing installation
xcodes requires superuser privileges in order to finish installation.
macOS User Password: 

Xcode 14.0.1 has been installed to /Applications/Xcode-14.0.1.app
```

Sources/xcodes/main.swift
MattKiazyk: The new `select` flag on the install is `"Select the installed xcode version after installation."` which its not doing. 

You're checking if the select flag is passed in, trying to select the new xcode (that isn't installed) and then installing it.

tahirmt: I was thinking of a use case where on CI you want to ensure that a certain Xcode version is installed. So you could do `xcodes install x.x.x --select --update`. If it's already installed you would expect it to be selected. 

So with all of these options if we try to select first, we exit super early by just selecting the version. But if we don't, it will first update the list, then see that it is already installed and select it. The list update will always happen even if Xcode was already installed and it wasn't needed. If that's the behavior you prefer I can remove it.

Sources/xcodes/main.swift
MattKiazyk: ah, i see now.. can we refactor that a bit. 

Can we remove the `exitAfterInstall` property on the `installer.install`. It's a bit confusing on doing it there if in the end we are using that to skip the select

Instead can we pass the select option inside this function and the last step check that, selecting if user wants.

tahirmt: I refactored this.

Sources/xcodes/main.swift
```diff
@@ -217,26 +226,57 @@ struct Xcodes: ParsableCommand {
...
+                    return xcodeList.update(dataSource: globalDataSource.dataSource)
+                        .then { _ -> Promise<InstalledXcode> in
+                            installer.install(installation, dataSource: globalDataSource.dataSource, downloader: downloader, destination: destination, experimentalUnxip: experimentalUnxip, emptyTrash: emptyTrash, noSuperuser: noSuperuser)
+                        }
+                } else {
+                    return installer.install(installation, dataSource: globalDataSource.dataSource, downloader: downloader, destination: destination, experimentalUnxip: experimentalUnxip, emptyTrash: emptyTrash, noSuperuser: noSuperuser)
+                }
+            }
+            .then { xcode -> Promise<Void> in
+                select ? selectXcode(shouldPrint: print, pathOrVersion: xcode.path.string, directory: destination, fallbackToInteractive: false) : .init()
```
MattKiazyk, line 269: Sorry one more #picky - can we if else on this select instead of `? :`

tahirmt, line 269:
```suggestion
                if select {
                    return selectXcode(shouldPrint: print, pathOrVersion: xcode.path.string, directory: destination, fallbackToInteractive: false) 
                } else {
                    return .init()
                }
```

tahirmt (commented):
If else

MattKiazyk (approved):
Thanks for the changes and willingness to wait.

jokerputihboy (approved):
Tepelekek

PR #189, opened 2022-03-16
Fix broken unit tests due to new logs and Xcode version

tahirmt:
There were unit tests that broke due to recent changes. Fixing the broken fixture files and also changing the CI to use Xcode 13 as that is now required to build the package


MattKiazyk (approved):
@tahirmt 🙏  Thanks!

Issue #177, opened 2022-01-09
Allow non-superuser installation


juanjonol:
Hi! Thank you for making this super useful tool!

Currently `xcodes` requires superuser privileges to work. As far as I know, this is required to:

- Enable developer mode, calling `/usr/sbin/DevToolsSecurity -enable` (I'm not sure what's this is for).
- Approve Xcode’s license.
- Install Xcode’s components.

None of this seems like a hard requirement that should make `xcodes` fail (Xcode will do all this on its first run anyway), so I think adding a `--no-superuser` option to `xcodes install` would make sense.

Would you accept a pull request for this?

Thank you!

om-ha:
> Enable developer mode, calling /usr/sbin/DevToolsSecurity -enable (I'm not sure what's this is for)

It's need for hooking to another process when debugging.
- https://www.unix.com/man-page/osx/1/DevToolsSecurity/
- https://stackoverflow.com/questions/9132826/stop-developer-tools-access-needs-to-take-control-of-another-process-for-debugg

Otherwise you'd get the prompt shown in the stackoverflow answer, second link above.

However, this command should automatically be called as per the Xcode Developer Tools installer.
> When the Xcode developer tools are installed, the post-install script automatically runs this DevToolsSecurity tool to change the authorization policies

PR #207, opened 2022-08-14
Added `no-superuser` flag

juanjonol:
The steps that require superuser (root) permissions are optional, so we can add a flag to skip them. This also makes `xcodes` more interactive (we don't need to stop to ask the user for its credentials).

This closes  https://github.com/RobotsAndPencils/xcodes/issues/177.


MattKiazyk (approved):

MattKiazyk:
Thanks @juanjonol I like it!

PR #211, opened 2022-09-10
Added support to immediately delete Xcode and its XIPs

### LINKED ISSUE #185: Add the ability to delete the xip file completely
The default behavior of moving the xip file to the trash after it is installed makes sense but I think it would be beneficial to add an option to delete the file completely instead so we don't have to do the manual process of cleaning trash afterwards

juanjonol:
Added a new `--empty-trash` flag to `xcodes install` and `xcodes uninstall`, to completely delete Xcode's .xip or the app itself instead of keeping it on the user's Trash.

This closes #185.

### DIFF
--- a/Sources/XcodesKit/XcodeInstaller.swift
+++ b/Sources/XcodesKit/XcodeInstaller.swift
@@ -166,9 +166,9 @@ public final class XcodeInstaller {
         case aria2(Path)
     }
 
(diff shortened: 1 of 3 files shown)


MattKiazyk (approved):

MattKiazyk:
Thanks @juanjonol

PR #214, opened 2022-09-14
Update Makefile - refactor swift-build option

ahmetgeymen:
trying to fix warning message comes from `swift build` during installation with `brew`.

`"warning: '--build-path' option is deprecated; use '--scratch-path' instead"`

reason: https://github.com/apple/swift-package-manager/pull/4195

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ xcodes: $(SOURCES)
 		--configuration release \
 		-Xswiftc -Onone \
 		--disable-sandbox \
-		--build-path "$(BUILDDIR)" \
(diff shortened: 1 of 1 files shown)


MattKiazyk (approved):

MattKiazyk:
Thanks @ahmetgeymen

MattKiazyk:
@ahmetgeymen letting you know that I've reverted this change in main for now, because we're still building with Swift 5.6 at the moment and that option doesn't exist yet

ahmetgeymen:
Ok then, not problem.

PR #215, opened 2022-09-17
Add ability to download Xcode without logging in using XcodeRelease

MattKiazyk:
I hate having to log into my Apple Developer account, always having to put in my 2fa code, when I want to download Xcode, but Apple has never given us the ability to do so.

As a CI user, it sucks that I have to type in my credentials all the time, while trying to do a xcode install on multiple machines. 

This PR changes that.

By using a public URL provided by Apple, we can get a token cookie that can be used to download the version of Xcode that we wish, without having to provide a log in. 🎉 

Caveats:
- This does not work when using the `Apple` datasource. More specifically that token doesn't work for listing the Xcode versions with `https://developer.apple.com/services-account/QH65B2/downloadws/listDownloads.action`

### DIFF
--- a/Sources/XcodesKit/Models.swift
+++ b/Sources/XcodesKit/Models.swift
@@ -50,6 +50,10 @@ public struct Xcode: Codable, Equatable {
     public let filename: String
     public let releaseDate: Date?
 
+    public var downloadPath: String {
+        return url.path
+    }
+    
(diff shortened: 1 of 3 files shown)


hardhub:
@MattKiazyk Does it still work? CLI asks me for Apple ID and if I do the _ADCDownloadAuthCookie_ trick manually with curl it redirects me to:
```
Location: https://developer.apple.com/unauthorized/
```

PR #220, opened 2022-09-21
Add "version" argument to "installed" function, printing Xcode's installation path

### LINKED ISSUE #219: Feature Request: Add ability to print the Xcode path of a given version
Hey 👋 

In order to support _fastlane_'s `xcversion` behavior of selecting a specific Xcode version **just during the lifecycle of the current _fastlane_ lane**, we could use some tooling that enabled `xcodes` to print the Xcode path for a given version.

The related code in _fastlane_ can be found here: https://github.com/fastlane/fastlane/blob/272063487b02b52510bb83172827fcbc336f4a55/fastlane/lib/fastlane/actions/xcversion.rb#L9 - as you can see, `xcode-install` (aka `xcversion`) has a function that returns an instance of "xcode" and sets its path as being the current DEVELOPER_DIR to be used by _fastlane_. If we had something similar in `xcodes`, then users wouldn't have to enter the sudo password at all when _temporarily_ selecting the Xcode version just for the current build.

rogerluan:
## Description

Resolves #219

This is another initiative that powers https://github.com/xcpretty/xcode-install/issues/467

With the resolution of #219, we can have a smoother transition from `xcode-install` to `xcodes`, including on _fastlane_ integrations.

See also: https://github.com/fastlane/fastlane/pull/20672

### DIFF
--- a/Sources/XcodesKit/XcodeInstaller.swift
+++ b/Sources/XcodesKit/XcodeInstaller.swift
@@ -762,6 +762,21 @@ public final class XcodeInstaller {
             }
     }
 
+    public func printXcodePath(ofVersion versionString: String, searchingIn directory: Path) -> Promise<Void> {
+        return firstly { () -> Promise<Void> in
+            guard let version = Version(xcodeVersion: versionString) else {
+                throw Error.invalidVersion(versionString)
+            }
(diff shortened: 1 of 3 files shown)


MattKiazyk (approved):
👍 Looks awesome - Thanks @rogerluan

PR #227, opened 2022-10-03
Use mocks when writing the cache file in unit tests

rpendleton:
After running the unit tests while working on #226, I noticed `xcodes list` was returning fake versions of Xcode. It looks like this is actually a pre-existing issue with the unit tests, so I figured I'd open a separate PR to fix it.


MattKiazyk (approved):
Thanks for the update!

PR #249, opened 2022-11-22
Better Handling of Apple Unauthorized errors

dmytro-kasianchuk-doordash:
If a new Apple ID account doesn't accept Developer Agreement at https://developer.apple.com/ then Xcode download will fail with Unauthorized error

However, the way Apple handles unauthorized errors is by redirecting to https://developer.apple.com/unauthorized/ and happily returning 200. Therefore the only way to check for authorization is to catch this redirect. If we don't do this, then unauthorized html will be downloaded and attempted to be unxipped, so the xcodes will fail with a "corrupted xip" error.

Therefore this PR:
1. Attempts to catch unauthorized redirect and throw error
2. Adds a descriptive message when this error is thrown
3. Prevents retries in case of 403. Retrying won't help, so it's better to terminate early

How to test:
1. In the `XcodeInstaller.swift` replace all instances of `sessionService.loginIfNeeded()` with `Promise()`
2. `xcodes signout`
3. Try to download e.g. `xcodes install 14.0`

Before the change you would get "corrupted xip message", after the change you would get a nice descriptive message.

It's all very hacky, I would love to hear suggestions on how to improve the PR

### DIFF
--- a/Sources/XcodesKit/Downloader.swift
+++ b/Sources/XcodesKit/Downloader.swift
@@ -49,7 +49,15 @@ public enum Downloader {
                                                                    to: destination.url,
                                                                    resumingWith: resumeData ?? persistedResumeData)
             progressChanged(progress)
-            return promise.map { $0.saveLocation }
+            return promise.map { result in
+                /// If the operation is unauthorized, the download page redirects to https://developer.apple.com/unauthorized/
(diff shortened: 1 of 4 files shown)


MattKiazyk (approved):
👍 

I've tested locally with overriding what gets returned and things seemed to be getting caught. I wish there was a simpler way (and there probably is) but this is an awesome addition to stop the xip file being an html file!

Thanks!

PR #230, opened 2022-10-07
Support runtime downloading/installing

StevenMagdy:
This PR builds on #223, and adds a command for installing runtimes `xcodes runtimes install 'iOS 14.5'`

~~Downloading will require an Apple account for new platforms released in 2022, it's not needed for older platforms.~~

Installing is still a WIP

PR #252, opened 2022-12-04
Unify `sudo` behavior for installing Xcode and runtimes

StevenSorial:
This PR removes the old `sudo` implementation, and uses the same behavior used by `runtimes` .


StevenSorial:
@StevenSorial I hope this PR can be merged soon 🤞🏼

StevenSorial:
@MattKiazyk Is this pr a no-go? I was hoping it could be merged, so follow it with other PRs.

StevenSorial:
Thank you for your response and sorry for my late reply. I will present some arguments why I think that might not be the right decision but feel free to skip them if it's not open for discussion.

### Consistency with `runtimes` 

#230 was accepted with this `sudo` implementation, so this PR just make it consistent across the whole app. It was not a security/stability concern for the runtimes command, so I don’t see why it will be for the other commands.


### Not a real concern

> DECISIONS.md
> * Run xcodes with sudo: xcodes doesn’t need elevated privileges the entire time, and means simple programmer mistakes can have bad consequences

>  It doesn’t allow some accidental code that needs higher permission from actually getting it and It’s being a nice player in open source.

While this concern is legitimate, I think it’s exaggerated. Many of the command line tools we use sometimes require sudo, and we don’t think much about it because they most probably have a good reason why they needed it, and because the community already tried and trusts the developers.

Also, In this PR and in #230, sudo is not always required, and only requested when the user entered a command that will require it, e.g,  for runtimes, sudo is not required when only downloading an image or when installing >=iOS 16 image.

Personally, as a user, I would be more concerned if a program is handling my password manually instead of the shell/system

### Hacky (IMO)

> It’s being a nice player in open source

Maybe I don’t look into many CLI tools code, but I personally have never seen this solution. Moreover, I see it as a maintainability burden.
I also had some problems with it while implementing #230 where it sometimes asked for the password while I already had privilege + It doesn’t work when using [touchid for sudo](https://sixcolors.com/post/2020/11/quick-tip-enable-touch-id-for-sudo/).

@MattKiazyk @rogerluan What do you think? I hope that I'm not arguing too much 😅

PR #257, opened 2022-12-13
Add support for Apple API sessions from Fastlane Spaceship

adamprice:
As I use fastlane and xcodes, I thought it could be cool if xcodes could use the same Apple API auth session as fastlane to avoid having to authenticate twice.

I've added two new command line options:

```bash
--use-fastlane-auth # enable/disable feature, default off
--fastlane-user # fastlane spaceship user, uses FASTLANE_SESSION env var if none provided
```

With this mode enabled, cookies are copied from either the `FASTLANE_SESSION` environment variable or `~/.fastlane/spaceship/<user>/cookie` file into an ephemeral `URLSessionConfiguration` that's then used for creating a `URLSession` to send requests. It also works with the way cookies are copied into the request for `aria2`. Hopefully this should stop the cookies from conficting with any existing sessions present in xcodes and they wont persist after the application has exited.

The current implementation is obviously coupled with how fastlane handles its cookies. As there's also a fastlane action for xcodes (https://docs.fastlane.tools/actions/xcodes/), another option could be coming up with a tool agnostic format for the cookies that the fastlane action can provide to xcodes.


tahirmt:
This is such a great idea! Something I tinkered with in the past but used ruby to achieve it. I do hope this gets in.

Sources/XcodesKit/Environment.swift
```diff
@@ -204,6 +204,13 @@ public struct Shell {
 }
 
 public struct Files {
+
+    public var homeDirectoryForCurrentUser: () -> URL = { FileManager.default.homeDirectoryForCurrentUser }
```
StevenSorial, line 208: We already have `Path.environmentHome` in `Path+.swift`

dgilbert-foreflight (commented):
This is great! Is this ready or still being drafted?

adamprice:
I'll update in the next couple of days and then remove the draft status.

ykhandelwal913:
@adamprice @tahirmt Is this ready for merge? This will solve the issue with CI environment

Sources/XcodesKit/Environment.swift
```diff
@@ -204,6 +205,7 @@ public struct Shell {
 }
 
 public struct Files {
+
```
tahirmt, line 208: Can we revert this

tahirmt (approved):
Looks good to me but we need a review from @MattKiazyk

ykhandelwal913:
@MattKiazyk can you please review this. This is very much needed as we are planning to migrate from xcversion to xcodes in our ci environment. xcodes seems to be very fast.

MattKiazyk:
@ykhandelwal913 I'll try to find some time this weekend to get it through. Sorry for the delay

Sources/XcodesKit/FastlaneSessionManager.swift
```diff
@@ -0,0 +1,60 @@
...
+        case Constants.fastlaneSessionEnvVarName:
+            importFastlaneCookiesFromEnv()
+        default:
+            importFastlaneCookiesFromFile(fastlaneUser: fastlaneUser)
+        }
+    }
+
+    private func importFastlaneCookiesFromEnv() {
+        guard let cookieString = Current.shell.env(Constants.fastlaneSessionEnvVarName) else {
+            Current.logging.log("\(Constants.fastlaneSessionEnvVarName) not set")
```
MattKiazyk, line 29: #picky can we add `.red` on these logs - so it stands out in the console?

MattKiazyk (approved):
Thanks for the time and contribution @adamprice!

Just a couple of picky additions to make the error messages stand out in the console!

Thanks again - sorry for the delay!

ykhandelwal913:
@MattKiazyk thanks for looking into it.

ykhandelwal913:
@MattKiazyk can that be merged?

ykhandelwal913:
@MattKiazyk we have a big dependency on fastlane session to migrate to xcodes. Can this PR request be merge if all checks are ok?

MattKiazyk:
Thanks @adamprice for all the work! 

@ykhandelwal913 this will be in the next version. However, I don't have a specific timeline on when that will be.

PR #261, opened 2023-01-04
Prevent keyNotFound error with unknown simulators

dnicolson:
This pull request prevents the following error when running `xcodes runtimes` when an "Unknown Platform Simulator" is installed:
```
Error: keyNotFound(CodingKeys(stringValue: "build", intValue: nil), Swift.DecodingError.Context(codingPath: [_JSONKey(stringValue: "820783CB-C389-4006-B2D2-D063F3FD10A1", intValue: nil)], debugDescription: "No value associated with key CodingKeys(stringValue: \"build\", intValue: nil) (\"build\").", underlyingError: nil))
```

If one accidentally runs `xcrun simctl runtime add` with an incompatible pre-iOS 16 Simulator runtime:
```
xcrun simctl runtime add com.apple.pkg.iPhoneSimulatorSDK15_0-15.0.1.1633542405.dmg
```

The following error will be displayed:
```
D: 94DEF0B1-1585-4D71-AB37-98AB08FDD000 <unknown platform> (Unusable - Missing Signature: Error Domain=SimDiskImageErrorDomain Code=3 "Missing Signature" UserInfo={NSLocalizedDescription=Missing Signature, unusableErrorDetail=})
```

The problem is that it's not immediately obvious that Xcode has still installed the runtime, but with a  missing signature state:
PixelSnap 2023-01-04 at 07 50 16@2x

And when xcodes runs `xcrun simctl runtime list -j`, there is no `build` key resulting in the error:
```
"820783CB-C389-4006-B2D2-D063F3FD10A1" : {
  "deletable" : true,
  "identifier" : "820783CB-C389-4006-B2D2-D063F3FD10A1",
  "kind" : "Disk Image",
  "path" : "\/Library\/Developer\/CoreSimulator\/Images\/Inbox\/820783CB-C389-4006-B2D2-D063F3FD10A1.dmg",
  "signatureState" : "Missing",
  "sizeBytes" : 5304795932,
  "state" : "Unusable",
  "unusableErrorDetail" : "Error Domain=SimDiskImageErrorDomain Code=3 \"Missing Signature\" UserInfo={NSLocalizedDescription=Missing Signature, unusableErrorDetail=}",
  "unusableErrorMessage" : "Missing Signature",
  "unusableSubstate" : "Missing Signature"
}
```

This change results in the following `xcodes runtimes` output:
```
-- Unknown --
Unknown 0 (Downloaded)
```

Another approach could be to filter out undecodable runtimes in the `RuntimeList` class, this actually resulted in more code though.

### DIFF
--- a/Sources/XcodesKit/Models+Runtimes.swift
+++ b/Sources/XcodesKit/Models+Runtimes.swift
@@ -86,6 +86,7 @@ extension DownloadableRuntime {
         case macOS = "com.apple.platform.macosx"
         case watchOS = "com.apple.platform.watchos"
         case tvOS = "com.apple.platform.appletvos"
+        case unknown = ""
 
         var order: Int {
             switch self {
@@ -93,6 +94,7 @@ extension DownloadableRuntime {
                 case .macOS: return 2
                 case .watchOS: return 3
                 case .tvOS: return 4
+                case .unknown: return 5
             }
         }
 
@@ -102,24 +104,25 @@ extension DownloadableRuntime {
                 case .macOS: return "macOS"
                 case .watchOS: return "watchOS"
                 case .tvOS: return "tvOS"
+                case .unknown: return "Unknown"
             }
         }
     }
 }
 
 public struct InstalledRuntime: Decodable {
-    let build: String
+    let build: String?
     let deletable: Bool
     let identifier: UUID
     let kind: Kind
     let lastUsedAt: Date?
     let path: String
-    let platformIdentifier: Platform
-    let runtimeBundlePath: String
-    let runtimeIdentifier: String
+    let platformIdentifier: Platform?
+    let runtimeBundlePath: String?
+    let runtimeIdentifier: String?
     let signatureState: String
     let state: String
-    let version: String
+    let version: String?
     let sizeBytes: Int?
 }
 
--- a/Sources/XcodesKit/RuntimeInstaller.swift
+++ b/Sources/XcodesKit/RuntimeInstaller.swift
@@ -43,13 +43,13 @@ public class RuntimeInstaller {
 
         installed.forEach { runtime in
             let resolvedBetaNumber = downloadablesResponse.sdkToSeedMappings.first {
-                $0.buildUpdate == runtime.build
+                $0.buildUpdate == runtime.build ?? ""
             }?.seedNumber
 
-            var result = PrintableRuntime(platform: runtime.platformIdentifier.asPlatformOS,
+            var result = PrintableRuntime(platform: runtime.platformIdentifier?.asPlatformOS ?? .unknown,
                                           betaNumber: resolvedBetaNumber,
-                                          version: runtime.version,
-                                          build: runtime.build,
+                                          version: runtime.version ?? "0",
+                                          build: runtime.build ?? "",
                                           state: runtime.kind)
 
             mappedRuntimes.indices {


StevenSorial:
Nice catch, thank you :)

> Another approach could be to filter out undecodable runtimes in the RuntimeList class, this actually resulted in more code though.

I prefer that approach. it doesn't have to be a lot of code, actually, it's less code.

Remove all the installed runtimes with the state `unusable`, before looping over them, then you can then force-unwrap `build`, `version`, and `platformIdentifier` (and remove the unknown enum case).

I would also add a note at the end with how many unknown runtimes were found, with a suggestion to run `xcrun simctl runtime list -j` for more info.

Disclaimer: I contributed the runtimes feature but I'm not part of the team, so you might want to wait for their opinion/review.

dnicolson:
The amount of unknown runtimes is indicated by how many "Unknown" runtimes are listed, it should be comparable to what is in Xcode. It's not ideal that there is a `0` though, but further hacks wouldn't have been nice.

The original approach seemed to require using a custom decoder and possibly filtering out the items unable to be decoded afterward. Or do you mean keeping the optional keys so that the decoding doesn't fail?

StevenSorial:
> Or do you mean keeping the optional keys so that the decoding doesn't fail?

yeah, what i meant is removing the
 ```
-- Unknown --
Unknown 0 (Downloaded)
```
 and replacing it at the end with (if more than 0): 
 ```
Note: Found 1 unknown downloaded runtime(s). For more info run `xcrun simctl runtime list -j`.
```

you just have to add `let unusableCount = installed.removeAll { $0.state == "Unusable" }.count` before looping over the `installed` array, and then you can force-unwrap the newly optional values.

Sources/XcodesKit/RuntimeInstaller.swift
```diff
@@ -38,18 +38,16 @@ public class RuntimeInstaller {
...
         }
 
-
-
-
+        let unusableCount = installed.removeAll { $0.state == "Unusable" }.count
         installed.forEach { runtime in
             let resolvedBetaNumber = downloadablesResponse.sdkToSeedMappings.first {
-                $0.buildUpdate == runtime.build
+                $0.buildUpdate == runtime.build!
```
MattKiazyk, line 44: Can we refactor this section a bit? 

Can we **not** use force unwrap. As a rule, I don't use them. Totally agree that it should _never_ happen but it's bitten me in the past. Perhaps making an overloaded `PrintableRuntime()` to allow optionals?

Sources/XcodesKit/Models+Runtimes.swift
```diff
@@ -108,18 +108,18 @@ extension DownloadableRuntime {
 }
 
 public struct InstalledRuntime: Decodable {
-    let build: String
+    let build: String?
```
MattKiazyk, line 111: #picky can we add a comment on why these are optional

Sources/XcodesKit/RuntimeInstaller.swift
dnicolson: I'm not sure how best to achieve this, feel free to make any changes.

StevenSorial: Just my 2 cents here, personally I also try to avoid them if I can, but in this case, I would prefer a force unwrap to fail than hiding the error and providing an empty string as a default value, in that case, we would know that `xcrun simctl runtime list` printed something that we didn't account for, and there is probably a new state of a runtime that we should write code for. I see it as a `assert`/`precondition` and not a code smell.

dnicolson: @MattKiazyk can you please give an update on how you would like to proceed here?

Bobbyphtr:
Hi, is there any updates for this PR?

dnicolson:
@MattKiazyk can you please give an update?

Issue #248, opened 2022-11-22
Automatic cookie based authentication


tahirmt:
Similar to how [fastlane](https://github.com/fastlane/fastlane) handles authentication cookies, it would be nice to be able to authenticate once for 30 days. For fastlane if you use `spaceauth` within 30 days with a stored cookie it doesn't ask for 2FA codes. This could help some people with maintaining CI. 

Along with this have a command `login` which just authenticates/reauthenticates with stored cookie and 2FA if needed. The cookie should be usable for further sessions. 

By doing something like this it would be only required to enter password once per 30 days.

triplef:
Yes please, this would be great to see added.

Issue #141, opened 2021-03-26
How to use in CI with 2FA enabled?


tahirmt:
We can use other things using fastlane and FASTLANE_SESSION to authenticate for CI. Is there a way to do that through environment variables?

senmu:
Thanks for opening this issue, @tahirmt! There are likely things from Fastlane we can leverage to automate this, but to fully understand the motivation behind this, can you describe how xcodes plays in your CI environment? Is this lack of support for 2FA currently blocking you right now?

tahirmt:
@senmu I'm using fastlane's xcode_install lane currently but xcodes is faster in terms of setting up an Xcode. We are managing our CI ourselves and I want to use xcodes in place of fastlane for the Xcode setup on the machine. No it is not a current blocker since there is an alternative we can use.

senmu:
Thanks for the insight, @tahirmt. At the moment our focus at Robots & Pencils is to ensure that the experience around manual use is taken care of first, so we won't have time to look into this in the near future. However if someone were to issue a PR for this, it would be welcomed. 

In the meantime, we'll keep it in the back of our minds and continue to think about how it can be integrated in such a way that it's done in a secure manner for when that time comes.

fkorotkov:
I think a viable workaround for the issue can be #145. One can manually put an Xcode XIP in the same network as CI nodes and install it from there.

FYI it seems it's now not possible to download Xcode without 2FA and `xcode-install` has the same issue. It might be a good opportunity for `xcodes` to be the only tool that can be used in CIs.

tejassharma96:
would also love to see an option for an install that requires no manual input (ie by using environment variables to represent the developer email, password, and session)

tahirmt:
I found a way to do this by making xcodes use the cookie generated by `fastlane` `spaceauth`. This allows for one login per 30 days. Which is how I'm managing our CI right now.
1. Use `fastlane spaceauth -u 'email'` to generate a fastlane cookie
2. If cookie isn't expired, run `fastlane spaceauth -u 'email'` again when reading cookies. (This is crucial because while the main cookie is valid, another cookie is only valid for 24 hours but running this doesn't require 2FA code. This bumps up the cookie time from 24 hours to 30 days.
2. Using a ruby script to load the cookies from the file stored by `spaceauth` into a json file that swift can read.
3. Load up the cookies into `HTTPCookieStorage` for `URLSession` to use.
[...]

PR #265, opened 2023-01-10
Added support for fastlane auth cookies

omarzl:
Solves issue https://github.com/RobotsAndPencils/xcodes/issues/141
Inspired by this fork commit: https://github.com/tahirmt/xcodes/commit/2da7761d34ace622044b2780b971d4c0c06b83d8

fastlane spaceauth generates a YML session string so we can parse it directly and inject those cookies into the session's cookies storage.

I kept the `FASTLANE_SESSION` env var name to have direct support with fastlane.

**Steps**

You will now be able to download Xcode in CI environments by first running locally:

```
fastlane spaceauth -u username@email.com
```

And then in your CI scripts you can export the variable and run `xcodes`, it won't prompt for username/password/2fa:

```
export  FASTLANE_SESSION='---\n-.......'
xcodes download 14.0
```

References:
https://github.com/fastlane/fastlane/blob/master/spaceship/lib/spaceship/client.rb#L280
https://github.com/sparklemotion/http-cookie/blob/master/lib/http/cookie_jar/yaml_saver.rb#L19

### DIFF
--- a/Sources/XcodesKit/FastlaneCookieLoader.swift
+++ b/Sources/XcodesKit/FastlaneCookieLoader.swift
@@ -1,0 +1,60 @@
+//
+//  FastlaneCookieLoader.swift
+//  
+//
+//  Created by Omar Zuñiga on 09/01/23.
+//
+
+import Foundation
+import Yams
+
+struct FastlaneCookieLoader {
+    
+    private let fastlaneSession = "FASTLANE_SESSION"
+    
+    func load(in storage: HTTPCookieStorage?) {
+        do {
+            guard let storage, var sessionVar = Current.shell.env(fastlaneSession) else { return }
+            // We need to convert literal line break into actual ones
+            sessionVar = sessionVar.replacingOccurrences(of: "\\n", with: "\n")
+            let decoder = YAMLDecoder()
+            let cookies: [Cookie] = try decoder.decode(from: sessionVar)
+            cookies.forEach { add(cookie: $0, in: storage) }
+        } catch {
+          print("Error decoding Fastlane cookie: \(error)")
+        }
+    }
+}
+
+private extension FastlaneCookieLoader {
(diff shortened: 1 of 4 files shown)


Sources/XcodesKit/FastlaneCookieLoader.swift
```diff
@@ -0,0 +1,60 @@
...
+import Foundation
+import Yams
+
+struct FastlaneCookieLoader {
+    
+    private let fastlaneSession = "FASTLANE_SESSION"
+    
+    func load(in storage: HTTPCookieStorage?) {
+        do {
+            guard let storage, var sessionVar = Current.shell.env(fastlaneSession) else { return }
```
tahirmt, line 17: I think it's also valuable to automatically load the cookie if it is present in the filesystem. fastlane stores it in the file `~/.fastlane/spaceship/email@domain.com/cookie`.

Sources/XcodesKit/FastlaneCookieLoader.swift
```diff
@@ -0,0 +1,60 @@
...
+//  FastlaneCookieLoader.swift
+//  
+//
+//  Created by Omar Zuñiga on 09/01/23.
+//
+
+import Foundation
+import Yams
+
+struct FastlaneCookieLoader {
```
tahirmt, line 11: Do you mind adding some tests for this?

tahirmt:
This will close https://github.com/RobotsAndPencils/xcodes/issues/248

tahirmt:
Also possible duplicate of https://github.com/RobotsAndPencils/xcodes/pull/257

omarzl:
> Also possible duplicate of #257

Oh there was another pr already parsing the cookie using Yams, I think we should prefer #257 instead of mine since it has a more complete solution

tahirmt:
@omarzl can you close this PR if it isn't needed?

omarzl:
Sure!

PR #267, opened 2023-01-15
Add a download runtimes command

### LINKED ISSUE #262: [feature] `download` subcommand for `runtimes`
It would really useful if it were possible to download runtimes without actually installing them. Like we can with Xcode versions.

i.e. `xcodes runtimes install "iOS 15.5"`

I would imagine this command to mount the DMG and extract the PKG to `~/Downloads`.

StevenSorial:
Resolves #262.

Also changed the word "Downloaded" to "Installed" in the runtimes list to avoid confusion.

### DIFF
--- a/Sources/XcodesKit/Downloader.swift
+++ b/Sources/XcodesKit/Downloader.swift
@@ -7,6 +7,14 @@ public enum Downloader {
     case urlSession
     case aria2(Path)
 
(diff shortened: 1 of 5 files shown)


MattKiazyk (approved):
Thanks @StevenSorial!

PR #276, opened 2023-03-04
Implement hashcash for AppleID Authentication

MattKiazyk:
Apple requires the `X-APPLE-HC` header now when signing in to https://idmsa.apple.com/appleauth/auth/signin. 

Leaving out this header results in Apple ID's possibly getting locked out, where the only way to get them back is to rest the Apple ID Password

Thanks to [arielmichaeli](https://twitter.com/arielmichaeli) for the figuring out the tough parts!

### DIFF
--- a/Sources/AppleAPI/URLRequest+Apple.swift
+++ b/Sources/AppleAPI/URLRequest+Apple.swift
@@ -2,6 +2,7 @@ import Foundation
 
 extension URL {
     static let itcServiceKey = URL(string: "https://appstoreconnect.apple.com/olympus/v1/app/config?hostname=itunesconnect.apple.com")!
+    // uses the get on signin to get the appropriate headers
     static let signIn = URL(string: "https://idmsa.apple.com/appleauth/auth/signin")!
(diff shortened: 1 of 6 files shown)


Sources/AppleAPI/URLRequest+Apple.swift
```diff
@@ -117,4 +119,14 @@ extension URLRequest {
...
+
+    /// Federate the sign in to get the X-Apple-HC header keys in order to properly mint a hashcash during regular signin
+    static func federate(account: String, serviceKey: String) throws -> URLRequest {
+        var request = URLRequest(url: .signIn)
+        request.allHTTPHeaderFields?["Accept"] = "application/json"
+        request.allHTTPHeaderFields?["Content-Type"] = "application/json"
+        request.httpMethod = "GET"
+        
+        return request
+    }
```
thispsj, line 131: @MattKiazyk can you explain me what would the Request look like in this case ?
Would the account and serviceKey be included as query parameters ?

PR #284, opened 2023-04-01
Converting all top-level commands to async

StevenSorial:
Converting all top-level commands to `async` in preparation to also convert the implementations in a later PR


hi2gage:
Just curious if there is a reason that this PR was closed? I would love to help moving xcodes over to async await.

PR #307, opened 2023-07-10
Support parsing `.xcode-version` beta format

### LINKED ISSUE #306: `xcode installed` command doesn't parse .xcode-version beta format
When using the Xcode 15 beta you can update your `.xcode-version` file to reference version "15" or "15.0" and as long as you only have 1 beta installed it will resolve the correct path. However, if you have more that one beta installed it will return an error due to it finding multiple matches, which is the desired behavior. In order to disambiguate which version you are trying to reference you can specify a version like "15.0b3". The issue is that the `xcode installed` command doesn't recognize this format.

This causes downstream issues when using tools like Fastlane for builds as it is able to properly determine the desired version, but then causes an error when querying xcodes.

Desired behavior:
`xcodes installed "15.0b2"`
`/Applications/Xcode-15.0.0-Beta.2.app`

Actual behavior:
`xcodes installed "15.0b2"`
`15.0 B 2 is not installed.`

rzulkoski:
Fixes #306

### DIFF
--- a/Sources/XcodesKit/Version+Xcode.swift
+++ b/Sources/XcodesKit/Version+Xcode.swift
@@ -43,8 +43,8 @@ public extension Version {
 
                     optionalPrereleaseIdentifiers[lastIndex] = "\(lastItem)-seed"
                 }
-			} else if type == "b" {
-				optionalPrereleaseIdentifiers.append("beta")
+            } else if type == "b" {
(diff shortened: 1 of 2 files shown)


MattKiazyk (approved):

MattKiazyk:
Thanks @rzulkoski! And sorry for the long response!

PR #318, opened 2023-09-21
Add dependabot support for swift

kemchenj:


MattKiazyk (approved):
Thanks!

I'm scared to see what this finds! If it's too much I might disable it later but lets see!

PR #389, opened 2024-10-20
Support Service Temporarily Unavailable response error

noppefoxwolf:
If Apple server response html format, Client can not decode response as JSON.
So, In this case, Client just look status code and throw serviceTemporarilyUnavailable error instead json decode error.

Ref: https://github.com/XcodesOrg/xcodes/issues/388
Ref: https://github.com/XcodesOrg/XcodesApp/issues/630


MattKiazyk:
👍 Thanks!

PR #393, opened 2024-10-29
Add support for SRP Apple login

### LINKED ISSUE #388: AppleID not recognized; Invalid JSON
Suddenly XCodes has an error reading my Account info. Worked yesterday, but this morning...hooped!
Checked my AppleID and all is good.
No Change in the information at all.
Looks like XCodes did an update during the night
Version: 2.3.0

CleanShot 2024-10-19 at 12 18 02

MattKiazyk:
Apple switched up their login to now use a Secure Remote Password type of algorithm. This add support for it, so a user can login properly.

Fixes #388

### DIFF
--- a/Sources/AppleAPI/Client.swift
+++ b/Sources/AppleAPI/Client.swift
@@ -2,6 +2,9 @@ import Foundation
 import PromiseKit
 import PMKFoundation
 import Rainbow
+import SRP
+import Crypto
+import CommonCrypto
 
 public class Client {
     private static let authTypes = ["sa", "hsa", "non-sa", "hsa2"]
@@ -20,6 +23,8 @@ public class Client {
         case invalidHashcash
         case missingSecurityCodeInfo
         case accountUsesHardwareKey
+        case srpInvalidPublicKey
+        case srpError(String)
         
         public var errorDescription: String? {
             switch self {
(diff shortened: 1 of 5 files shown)


Sources/XcodesKit/Environment.swift
```diff
@@ -300,7 +300,7 @@ public struct Network {
 
     public var validateSession: () -> Promise<Void> = client.validateSession
 
-    public var login: (String, String) -> Promise<Void> = { client.login(accountName: $0, password: $1) }
+    public var login: (String, String) -> Promise<Void> = { client.srpLogin(accountName: $0, password: $1) }
```
tsdevshop, line 303: Is it worth gating the old login approach behind a flag? This is the approach that [fastlane has taken](https://github.com/fastlane/fastlane/pull/26415).

tsdevshop (commented):
Is it possible to add any tests for this new flow? Similar to [those in the Xcodes.app repo](https://github.com/XcodesOrg/XcodesApp/pull/640/files#diff-4c3a6519a6878fb69eb4776e7697714c3e64dfde9349b0f0f271823d363f42590)

Sources/AppleAPI/Client.swift
```diff
@@ -56,6 +61,90 @@ public class Client {
             }
     }
     
+    public func srpLogin(accountName: String, password: String) -> Promise<Void> {
```
spadafiva, line 64: If this references an implementation or specific other specification, it would be nice to link it as part of a comment for what/how you followed to implement this

Sources/AppleAPI/Client.swift
```diff
@@ -56,6 +61,90 @@ public class Client {
...
+        var serviceKey: String!
+        let client = SRPClient(configuration: SRPConfiguration<SHA256>(.N2048))
+        let clientKeys = client.generateKeys()
+        let a = clientKeys.public
+        
+        return firstly { () -> Promise<(data: Data, response: URLResponse)> in
```
spadafiva, line 70: Non-blocking, but it would be good to have a small comment around each of the steps here since it seems like an involved handshake process

spadafiva (approved):
Fix is working for me and using the SRP login!  Thanks @MattKiazyk

Sources/XcodesKit/Environment.swift
MattKiazyk: I don't see Apple ever returning back to their old style login. Xcodes doesn't have to be as backwards compatible as fastlane does, so I'm ok taking it out.

PR #398, opened 2024-11-05
support legacy SRP (s2k_fo) logins

MattKiazyk:
Duplicate of https://github.com/XcodesOrg/XcodesApp/pull/650

> Apple IDs with "old" passwords (ie those set before a certain date, the cutoff for which I'm not 100% certain) use the s2k_fo protocol rather than s2k. This legacy protocol involved an extra step in the SRP handshake, where the password was hex-encoded before performing PBKDF2 to derive the shared secret. Luckily it's pretty trivial for us to handle this and fix the erroneous "incorrect username or password" errors that it resulted in.

### DIFF
--- a/Sources/AppleAPI/Client.swift
+++ b/Sources/AppleAPI/Client.swift
@@ -105,7 +105,7 @@ public class Client {
             let iterations = srpInit.iteration
             
             do {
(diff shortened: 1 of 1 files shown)


controversial:
What’s blocking this?

PR #429, opened 2025-06-10
use the correct org name

pastey:
Hi!

Noticed that the [Release description](https://github.com/XcodesOrg/xcodes/releases/tag/1.6.1) references the original `robotsandpencils` organization. Changed it to the `xcodesorg` in the corresponding github release template.

`grep -R robotsandpencils` shows some other places, where `robotsandpencils` is referenced. I'm sure that the one in the `CONTRIBUTING.md` should be updated.

As for the rest, I've left them untouched, as I'm afraid, they might require some migration code:
 - XcodesKitTests.swift
 - LogOutput-DamagedXIP.txt
 - Environment.swift
 - Path+.swift

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -11,7 +11,7 @@ categories:
       - 'documentation'
       - 'dependencies'
 template: |
-  If you installed xcodes with homebrew you can upgrade with `brew upgrade robotsandpencils/made/xcodes`.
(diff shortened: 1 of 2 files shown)


MattKiazyk:
Thanks @pastey i'm not too concerned about the name at the moment around the app but thanks for the PR.

PR #437, opened 2025-07-10
support multiple architectures for runtimes

MattKiazyk:
Starting in Xcode 26 beta 2, Apple is now returning different runtime builds for different architectures.

### Before
```
xcodes runtimes --include-betas

iOS 18.5 (Installed)
iOS 26.0-beta1 (Installed)
iOS 26.0-beta1 (Installed)
iOS 26.0-beta2
iOS 26.0-beta3
iOS 26.0
iOS 26.0
```
### After
```
xcodes runtimes --include-betas

iOS 18.5  (Installed)
iOS 26.0-beta1 arm64|x86_64 (Installed)
iOS 26.0-beta1 arm64|x86_64 (Installed)
iOS 26.0-beta2 arm64|x86_64
iOS 26.0-beta2 arm64
iOS 26.0-beta3 arm64|x86_64
iOS 26.0-beta3 arm64
```

to fully install now, users will have to use `iOS 26.0-beta3 arm64|x86_64` as the full identifier to distinguish the id as both have the same build version. 

Installing via build version will still work but (I believe) will grab the universal variant. 

NOTE: In this PR `xcodes runtimes install "iOS 26.0-beta3 arm64"` won't actually install the Apple Silicon variant. Need to add in support for `-architectureVariant <universal|arm64>` for the `xcodebuild` on Xcode 26 beta - logged in https://github.com/XcodesOrg/xcodes/issues/436

### DIFF
--- a/.swiftpm/xcode/xcshareddata/xcschemes/xcodes.xcscheme
+++ b/.swiftpm/xcode/xcshareddata/xcschemes/xcodes.xcscheme
@@ -128,8 +128,12 @@
             isEnabled = "NO">
          </CommandLineArgument>
          <CommandLineArgument
-            argument = "runtimes install &quot;visionOS 1.0-beta1&quot;"
+            argument = "runtimes install &quot;visionOS 26.0-beta3 arm64&quot;"
             isEnabled = "YES">
+         </CommandLineArgument>
+         <CommandLineArgument
+            argument = "runtimes --include-betas"
+            isEnabled = "NO">
          </CommandLineArgument>
          <CommandLineArgument
             argument = "install --help"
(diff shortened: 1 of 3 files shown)


icedice:
Who should review and merge this?

edit:

I thought this would solve beta5 not showing up but that does not seem to be the case. I tried building the branch but beta5 still does not show.

trentguillory-gc (approved):
Nice update. And important improvement. I don't have any sway here, but I approve of these changes. 

And #440 should be merged too.

gianpispi (approved):
I have no saying in this, but It looks good to me. Hopefully helping this to be merged since we can't use it right now.

trsneed:
I would really like this merged too!

PR #459, opened 2026-01-16
Handle multiple architectures: filter out architectures that dont match with machine's architecture

acecilia:
Solves https://github.com/XcodesOrg/xcodes/issues/456

Tested on my intel mac: after running `update`, I was able to run `install 26.2` and it installed the correct architecture

### DIFF
--- a/Sources/XcodesKit/XcodeList.swift
+++ b/Sources/XcodesKit/XcodeList.swift
@@ -23,9 +23,11 @@ public final class XcodeList {
     }
 
     public func update(dataSource: DataSource) -> Promise<[Xcode]> {
+        let xcodesPromise: Promise<[Xcode]>
+
         switch dataSource {
         case .apple:
-            return when(fulfilled: releasedXcodes(), prereleaseXcodes())
+            xcodesPromise = when(fulfilled: releasedXcodes(), prereleaseXcodes())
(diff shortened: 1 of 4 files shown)


Sources/XcodesKit/XcodeList.swift
```diff
@@ -190,5 +192,45 @@ extension XcodeList {
...
+                let selected: Xcode
+                if isAppleSilicon {
+                    // Prefer Apple_silicon, fallback to Universal
+                    selected = variants.first { $0.url.absoluteString.contains("Apple_silicon") }
+                            ?? variants.first { $0.url.absoluteString.contains("Universal") }
+                            ?? variants[0]
+                } else {
+                    // On Intel, prefer Universal, avoid Apple_silicon
+                    selected = variants.first { $0.url.absoluteString.contains("Universal") }
+                            ?? variants.first { !$0.url.absoluteString.contains("Apple_silicon") }
```
acecilia, line 226: Datasource does not include an `arch` parameter, so the only way to distinguish is by using the file name

acecilia:
@MattKiazyk I see you faced some similar issue with runtimes in [here](https://github.com/XcodesOrg/xcodes/pull/437). What do you think about this approach?

acecilia:
Ping @MattKiazyk 🙏

PR #461, opened 2026-02-11
README: Point user to `homebrew-core`, not the `xcodesorg/made` tap

DilumAluthge:
On the machine I tried (macOS Tahoe on Apple Silicon), bottles are not available with the [`xcodesorg/made` tap](https://github.com/XcodesOrg/homebrew-made), but they are available when using [`homebrew/core`](https://github.com/Homebrew/homebrew-core/blob/main/Formula/x/xcodes.rb).

This PR updates the README to point the users to the latter.

An alternative solution would be make bottles available in the `xcodesorg/made` tap (https://github.com/XcodesOrg/homebrew-made/issues/3).

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Xcodes is now part of the `XcodesOrg` - [read more here](nextstep.md)
 ### Homebrew (Preferred)
 
 ```sh
-brew install xcodesorg/made/xcodes
+brew install xcodes
 ```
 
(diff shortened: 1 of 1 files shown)


DilumAluthge:
@MattKiazyk Could you review this?

DilumAluthge:
Bump @MattKiazyk Would you be able to review this? If not, would you be able to suggest an alternate reviewer?

PR #481, opened 2026-06-16
Fix 2FA authentication regression causing 403 on download (#480)

### LINKED ISSUE #480: 403: Unauthorized received on accounts with two-factor enabled.
When trying downloading a version using the install command using a device that has two-factor enabled a 403: Unauthorized is shown.
The username & password are correctly entered and when running the command a two-factor code is shown on the device itself.
The terms & conditions are  setup correctly and there are no other weird restrictions on the account(s) (I've tested multiple).

```shell
evertwoud@evertwoud ~ % xcodes install 26.4.1

(1/6) Found existing archive that will be used for installation at /Users/evertwoud/Library/Application Support/com.robotsandpencils.xcodes/Xcode-26.4.1+17E202.xip.
Using regular unxip. Try passing `--experimental-unxip` for a faster unxip process
The archive "Xcode-26.4.1+17E202.xip" is damaged and can't be expanded.
Removing damaged XIP and re-attempting installation.


Downloading with aria2 (/opt/homebrew/bin/aria2c)

 Received 403: Unauthorized. This can happen when either:
 1. Apple Developer Terms and Conditions were not accepted at https://developer.apple.com/
 2. Apple ID authorization was revoked for some other reason
```

kinoroy:
## Summary

Fixes #480 — accounts with two-factor authentication get a `403: Unauthorized` on download because the CLI never completes the 2FA flow (no verification-code prompt appears).

## Root cause

In **1.6.x**, `Client.srpLogin` drove the entire interactive 2FA flow internally — on Apple's `409` response it prompted for the verification code, submitted it, trusted the session, and only returned once fully authenticated.

In **2.x**, login moved to the external [`XcodesLoginKit`](https://github.com/XcodesOrg/XcodesLoginKit) package and was redesigned as a state machine. `srpLogin` now **returns** an `AuthenticationState` (e.g. `.waitingForSecondFactor`) instead of completing the flow — the caller is expected to drive the second factor (the SwiftUI app does this). But the CLI's `login` closures in `Environment.swift` discarded the returned state:

```swift
_ = try await loginClient.srpLogin(accountName: accountName, password: password)
```

So for any 2FA account: `srpLogin` returned `.waitingForSecondFactor` without throwing → `AppleSessionService.loginIfNeeded` treated it as success → the cookie jar held only the *pre-2FA* session → the download hit Apple's **403 Unauthorized**. The missing 2FA prompt was the visible symptom; the 403 was the downstream consequence. Non-2FA accounts were unaffected, which matches the report that this is a regression specific to 2FA accounts.

## The fix

- **`TwoFactorAuthentication.swift`** (new): Ports 1.6.x's interactive completion onto the new state-machine API. `completeIfNeeded(_:dependencies:)` inspects the returned state and handles every path 1.6.x supported:
  - **Trusted-device code** (with `"sms"` fallback to phone-number selection)
  - **SMS auto-sent** to a single trusted number
  - **SMS phone-number selection** (lists numbers, requests the code, retries on invalid index)
  - **Security key** and **federated / not-developer** states surface clear errors (1.6.x never supported hardware keys either).

  Modeled with injected closures (`Dependencies`) rather than the concrete `Client`, matching the repo's existing dependency-injection style and keeping it unit-testable.
- **`Environment.swift`**: Both `login` closures now capture the returned state and call `TwoFactorAuthentication.completeIfNeeded(...)`.
- **`TwoFactorAuthenticationTests.swift`** (new): 8 tests covering each branch.

## Testing

- `swift test` — **87 passed, 0 failures** (was 79; +8 new).
- Manually verified `xcodes install` end-to-end against a real Apple ID with **SMS 2FA** and against one with **trusted-device 2FA** — both now prompt for the code and authenticate successfully.

### DIFF
--- a/Sources/XcodesKit/Environment.swift
+++ b/Sources/XcodesKit/Environment.swift
@@ -296,7 +296,8 @@ public struct Network: Sendable {
             downloadTask = { loginClient.urlSession.downloadTask(with: $0, to: $1, resumingWith: $2) }
             validateSession = { _ = try await loginClient.validateSession() }
             login = { accountName, password in
-                _ = try await loginClient.srpLogin(accountName: accountName, password: password)
+                let state = try await loginClient.srpLogin(accountName: accountName, password: password)
+                try await TwoFactorAuthentication.completeIfNeeded(state, dependencies: .liveDependencies(client: loginClient))
             }
             checkIsFederated = { accountName in
                 try await loginClient.checkIsFederated(accountName: accountName)
@@ -371,7 +372,8 @@ public struct Network: Sendable {
         self.downloadTask = downloadTask ?? { loginClient.urlSession.downloadTask(with: $0, to: $1, resumingWith: $2) }
         self.validateSession = validateSession ?? { _ = try await loginClient.validateSession() }
         self.login = login ?? { accountName, password in
-            _ = try await loginClient.srpLogin(accountName: accountName, password: password)
+            let state = try await loginClient.srpLogin(accountName: accountName, password: password)
+            try await TwoFactorAuthentication.completeIfNeeded(state, dependencies: .liveDependencies(client: loginClient))
         }
         self.checkIsFederated = checkIsFederated ?? { accountName in
             try await loginClient.checkIsFederated(accountName: accountName)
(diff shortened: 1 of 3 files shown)


MattKiazyk:
Thanks @kinoroy - I'll try to get this merged and built soon!

MattKiazyk (approved):

